PR #4, opened 2021-06-25
Media DAG

tomthornton:


dags/rock_media.py
```diff
@@ -0,0 +1,126 @@
...
+    )
+
+    def get_content_item_id(rockOriginId):
+        return pg_hook.get_first(f'SELECT id FROM "contentItems" WHERE "originId"::Integer = {rockOriginId}')[0]
+
+    def mapContentItems(contentItem):
+    
+        nodeId = get_content_item_id(contentItem['Id'])
+
+        def is_media_image( attribute ):
```
vinnyjth, line 29: 🙌

dags/rock_media.py
```diff
@@ -0,0 +1,126 @@
...
+                return 'AUDIO'
+            else:
+                return 'UNKNOWN_MEDIA_TYPE'
+
+        def get_media_value( attribute  ):
+            mediaType = get_media_type ( attribute )
+            attributeKey = attribute['Key']
+            attributeValue = contentItem['AttributeValues'][attributeKey]['Value']
+
+            if(mediaType == 'IMAGE'):
```
vinnyjth, line 62: Good catch here, thanks!

dags/backfill-rock-media-dag.py
```diff
@@ -0,0 +1,31 @@
+from airflow import DAG
```
vinnyjth, line 1: Rather than having this as a separate DAG, what would you think about having this run in series with the othetr operators in  `content_item`?

Something like 

`fetch content items -> [fetch_media, fetch_relationships]`

dags/test_rock_media.py
```diff
@@ -0,0 +1,117 @@
+import requests
```
vinnyjth, line 1: I like this test file :) I'm all for leaving this in here for reference / usage

dags/test_rock_media.py
```diff
@@ -0,0 +1,117 @@
...
+
+            }
+
+        filteredAttributes = filter(filter_media_attributes, contentItem['Attributes'].values())
+        mappedAttributes = map(map_attributes, filteredAttributes)
+
+        return list(mappedAttributes)
+
+def fetch_and_save_media():
+    headers = {"Authorization-Token": "ASZjZWdf3IqrbZX9sedtB4wb"}
```
vinnyjth, line 87: We should probably take the auth token out

vinnyjth (changes requested):
Looking good! A few pieces of feedback then we'll be ready to go. I'll merge this after we get the `originId/originType` PR merged in `apps`.

dags/test_rock_media.py
tomthornton: Yeah for sure. I actually didn't mean to commit this, I was just keeping it locally for reference.

dags/backfill-rock-media-dag.py
tomthornton: Definitely. I'll make those changes.

dags/backfill-rock-media-example-dag.py
```diff
@@ -0,0 +1,31 @@
+from airflow import DAG
```
vinnyjth, line 1: I think you can delete this file now

PR #5, opened 2021-06-28
Cover Image

tomthornton:
Adds cover image to content item


tomthornton:
@vinnyjth are you thinking this should be it's own DAG? I would guess so since the cover image could be updated

vinnyjth:
@tomthornton I would keep this as an additional operator in the `content_item` dag :) You can re-run just one of the operators out of a dag as needed, and the entire `dag` will be re-run frequently enough so I don't think it should be a problem.

PR #7, opened 2021-06-28
Content Item Category Operator

vinnyjth:
image
image

PR #8, opened 2021-06-28
Cover Image Dag

vinnyjth:
@tomthornton this PR is a bit weird, it's a "revert" of a "revert" but it's got all your changes

### DIFF
--- a/dags/rock_cover_image.py
+++ b/dags/rock_cover_image.py
@@ -80,7 +80,10 @@ def fetch_and_save_cover_image(ds, *args, **kwargs):
         if(args):
             contentItemId = args['ContentItemId']
             coverImageId = args['CoverImageId']
-            return pg_hook.run('UPDATE "contentItems" SET "coverImageId" = %s WHERE "originId"::Integer = %s', True, (coverImageId, contentItemId))
+            return pg_hook.run(
+                'UPDATE "contentItems" SET "coverImageId" = (SELECT id FROM "media" WHERE "originId" = %s) WHERE "originId"::Integer = %s', 
+                True, 
+                (coverImageId, contentItemId))
 
     while fetched_all == False:
         # Fetch people records from Rock.
--- a/dags/rock_cover_image.py
+++ b/dags/rock_cover_image.py
@@ -1,0 +1,105 @@
+from airflow.models import Variable
+from airflow.hooks.postgres_hook import PostgresHook
+import requests
+
+def get_images (content_item):
+    def filter_images( attribute ):
+            attributeKey = attribute['Key']
+            attributeValue = content_item['AttributeValues'][attributeKey]['Value']
+
+            return attribute['FieldTypeId'] == 10 or ('image' in attributeKey.lower() and isinstance(attributeValue, str) and attributeValue.startswith('http'))
+
+    return list(filter(filter_images, content_item['Attributes'].values()))
+
+def get_best_image_id (images):
+    if(len(images) > 1):
+        squareImages = list(filter(lambda attribute: 'square' in attribute['Key'].lower(), images ))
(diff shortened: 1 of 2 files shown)


dags/rock_cover_image.py
```diff
@@ -0,0 +1,105 @@
...
+            parents = requests.get(
+                    f"{Variable.get(kwargs['client'] + '_rock_api')}/ContentChannelItemAssociations",
+                    params={
+                        "$filter": f'ChildContentChannelItemId eq {contentItemId}',
+                    },
+                    headers=headers).json()
+
+            if(len(parents) == 1):
+                parentContentItemId = parents[0]['ContentChannelItemId']
+                parentContentItem = requests.get(
```
vinnyjth, line 67: Your approach here looks great! I am a bit worried about the number of queries we'll have to make to Rock per-content item. We have all the data we need in Postgres _except_ the `key` :/ What do you think about moving that key into postgres long term? Or maybe we split shoveling into two stages 

1.  rock -> postgres, we don't look at parents 
2. postgres -> postgres, we look for all items that don't have cover images and give them their parents cover image, purely by querying postgres.

dags/rock_cover_image.py
```diff
@@ -0,0 +1,105 @@
...
+                    "ContentItemId": content_item['Id'],
+                    "CoverImageId": coverImageId
+                }
+
+
+    def update_content_item_cover_image(args):
+        if(args):
+            contentItemId = args['ContentItemId']
+            coverImageId = args['CoverImageId']
+            return pg_hook.run('UPDATE "contentItems" SET "coverImageId" = %s WHERE "originId"::Integer = %s', True, (coverImageId, contentItemId))
```
vinnyjth, line 83: we talked about this in slack, but we should be setting `coverImageId` to the `media.id`

vinnyjth: https://github.com/ApollosProject/apollos-shovel/pull/7/files#diff-8b958333e85b50db625a8a72e3523e1ff545dd54cc97728257a758b9ae0c80f2R162-R166

Look here for an example of how you can still use the `originId` to set the `mediaId`

dags/rock_cover_image.py
```diff
@@ -0,0 +1,105 @@
...
+def get_images (content_item):
+    def filter_images( attribute ):
+            attributeKey = attribute['Key']
+            attributeValue = content_item['AttributeValues'][attributeKey]['Value']
+
+            return attribute['FieldTypeId'] == 10 or ('image' in attributeKey.lower() and isinstance(attributeValue, str) and attributeValue.startswith('http'))
+
+    return list(filter(filter_images, content_item['Attributes'].values()))
+
+def get_best_image_id (images):
```
vinnyjth, line 14: Nice work! you captured this logic well

vinnyjth:
@tomthornton Left some feedback above :)

dags/rock_cover_image.py
tomthornton: Yep! just added a commit for this

dags/rock_cover_image.py
tomthornton: Thanks!

tomthornton:
@vinnyjth updated to grab the right Id. As far as storing "key" long term goes, I think that would be fine but wouldn't it be better to store Parent -> Child relationships of the content items instead? I feel like that will be more useful in the long term.

vinnyjth:
@tomthornton What do you think about commenting out the "pull image from parent" logic for now, merging this, and then seeing if we can come up with a sql query solution to solve the parent logic?

tomthornton:
@vinnyjth Sounds good to me. Shouldn't be too hard.

PR #10, opened 2021-06-30
fix: Update content item's cover image to be parent's if null

tomthornton:
Fancy SQL query from @vinnyjth

### DIFF
--- a/dags/rock_cover_image.py
+++ b/dags/rock_cover_image.py
@@ -1,0 +1,113 @@
+from airflow.models import Variable
+from airflow.hooks.postgres_hook import PostgresHook
+import requests
+
+def get_images (content_item):
+    def filter_images( attribute ):
+            attributeKey = attribute['Key']
+            attributeValue = content_item['AttributeValues'][attributeKey]['Value']
+
+            return attribute['FieldTypeId'] == 10 or ('image' in attributeKey.lower() and isinstance(attributeValue, str) and attributeValue.startswith('http'))
+
+    return list(filter(filter_images, content_item['Attributes'].values()))
+
+
+
+def get_best_image_id (images, pg_hook):
+    imageId = None;
+    if(len(images) > 1):
+        squareImages = list(filter(lambda attribute: 'square' in attribute['Key'].lower(), images ))
+        if(len(squareImages) > 0):
+            imageId = squareImages[0]['Guid']
+        else:
+            return images[0]['Guid']
+    elif(len(images) == 1):
+        imageId = images[0]['Guid']
+
+    if(imageId):
+        print(str(imageId))
+        return pg_hook.get_first('SELECT id FROM "media" WHERE "originId" = %s', (imageId,))[0]
+    
+    return None
+
+
+
+
+def fetch_and_save_cover_image(ds, *args, **kwargs):
+    if 'client' not in kwargs or kwargs['client'] is None:
+        raise Exception("You must configure a client for this operator")
+    
+    headers = {"Authorization-Token": Variable.get(kwargs['client'] + "_rock_token")}
+
+    pg_connection = kwargs['client'] + '_apollos_postgres'
+    pg_hook = PostgresHook(postgres_conn_id=pg_connection,
+        keepalives=1,
+        keepalives_idle=30,
+        keepalives_interval=10,
+        keepalives_count=5
+    )
+
+    fetched_all = False
+    skip = 0
+    top = 10000
+
+    def map_content_items(content_item):
+
+        imageAttributes = get_images(content_item)
+        coverImageId = get_best_image_id(imageAttributes, pg_hook)
+
+        if(coverImageId):
+            return {
+                "ContentItemId": content_item['Id'],
+                "CoverImageId": coverImageId
+            }
+        else:
+            contentItemId = content_item['Id']
+
+            parents = requests.get(
+                    f"{Variable.get(kwargs['client'] + '_rock_api')}/ContentChannelItemAssociations",
+                    params={
+                        "$filter": f'ChildContentChannelItemId eq {contentItemId}',
+                    },
+                    headers=headers).json()
+
+            if(len(parents) == 1):
+                parentContentItemId = parents[0]['ContentChannelItemId']
+                parentContentItem = requests.get(
+                    f"{Variable.get(kwargs['client'] + '_rock_api')}/ContentChannelItems/{parentContentItemId}",
+                    params={"loadAttributes": "expanded"},
+                    headers=headers).json()
+                coverImageId = get_best_image_id(get_images(parentContentItem), pg_hook)
+                
+                return {
+                    "ContentItemId": content_item['Id'],
+                    "CoverImageId": coverImageId
+                }
+
+    def update_content_item_cover_image(args):
+        if(args):
+            contentItemId = args['ContentItemId']
+            coverImageId = args['CoverImageId']
+            return pg_hook.run('UPDATE "contentItems" SET "coverImageId" = %s WHERE "originId"::Integer = %s', True, (coverImageId, contentItemId))
+
+    while fetched_all == False:
+        # Fetch people records from Rock.
+
+        params = {
+            "$top": top,
+            "$skip": skip,
+            # "$select": "Id,Content",
+            "loadAttributes": "expanded",
+            "$orderby": "ModifiedDateTime desc",
+        }
+
+        r = requests.get(
+                f"{Variable.get(kwargs['client'] + '_rock_api')}/ContentChannelItems",
+                params=params,
+                headers=headers)
+                
+        contentItems = list(map(map_content_items, r.json()))
+        updatedItems = list(map(update_content_item_cover_image, contentItems))
+
+        skip += top
+        fetched_all = len(r.json()) < top
--- a/dags/rock_cover_image.py
+++ b/dags/rock_cover_image.py
@@ -51,15 +51,24 @@ def fetch_and_save_cover_image(ds, *args, **kwargs):
     skip = 0
     top = 10000
 
+    def update_content_item_cover_image(args):
+        contentItemId = args['ContentItemId']
+        coverImageId = args['CoverImageId']
+        return pg_hook.run(
+            'UPDATE "contentItems" SET "coverImageId" = (SELECT id FROM "media" WHERE "originId" = %s) WHERE "originId"::Integer = %s', 
+            True, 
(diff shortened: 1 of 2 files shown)


dags/rock_cover_image.py
```diff
@@ -6,28 +6,35 @@ def get_images (content_item):
...
-def get_best_image_id (images):
+
+
+def get_best_image_id (images, pg_hook):
+    imageId = None;
     if(len(images) > 1):
         squareImages = list(filter(lambda attribute: 'square' in attribute['Key'].lower(), images ))
         if(len(squareImages) > 0):
-            return squareImages[0]['Id']
+            imageId = squareImages[0]['Guid']
```
vinnyjth, line 21: So - one thing I ran into earlier is that `Guid` is the same for every attribute of the same type. So every content item of the same type has the same image attribute, but the values of those attributes change. 

For that reason - as you can see from the div, I ended up making a new originId by combining the `attribute.id` and the `content_item.id` 
`str(content_item['Id']) + "/" + str(attribute[id])`

We'll need to do that here - or at least do it on line 29

dags/rock_cover_image.py
```diff
@@ -6,28 +6,35 @@ def get_images (content_item):
     def filter_images( attribute ):
             attributeKey = attribute['Key']
             attributeValue = content_item['AttributeValues'][attributeKey]['Value']
-
-            return (attribute['FieldTypeId'] == 10 and attributeValue) or ('image' in attributeKey.lower() and isinstance(attributeValue, str) and attributeValue.startswith('http'))
+          
+            return attribute['FieldTypeId'] == 10 or ('image' in attributeKey.lower() and isinstance(attributeValue, str) and attributeValue.startswith('http'))
```
vinnyjth, line 10: We need to revert this change :) this is to help prevent returning values without an image.

dags/rock_cover_image.py
```diff
@@ -6,28 +6,35 @@ def get_images (content_item):
...
+            return images[0]['Guid']
     elif(len(images) == 1):
-        return images[0]['Id']
-    else:
-        return None
+        imageId = images[0]['Guid']
+
+    if(imageId):
+        print(str(imageId))
+        return pg_hook.get_first('SELECT id FROM "media" WHERE "originId" = %s', (imageId,))[0]
```
vinnyjth, line 29: awesome change here. returning the actual postgres ID certainly makes this logic easier to follow.

dags/rock_cover_image.py
```diff
@@ -42,51 +49,24 @@ def fetch_and_save_cover_image(ds, *args, **kwargs):
...
+        return pg_hook.run(
+            'UPDATE "contentItems" SET "coverImageId" = (SELECT id FROM "media" WHERE "originId" = %s) WHERE "originId"::Integer = %s', 
+            True, 
+            (coverImageId, contentItemId))
+
     def map_content_items(content_item):
 
         imageAttributes = get_images(content_item)
-        coverImageId = get_best_image_id(imageAttributes)
+        coverImageId = get_best_image_id(imageAttributes, pg_hook)
```
vinnyjth, line 63: if you change where `get_best_image_id` is defined, you can avoid passing `pg_hook` :D

dags/rock_cover_image.py
```diff
@@ -103,9 +83,26 @@ def update_content_item_cover_image(args):
                 f"{Variable.get(kwargs['client'] + '_rock_api')}/ContentChannelItems",
                 params=params,
                 headers=headers)
-
+                
         contentItems = list(map(map_content_items, r.json()))
-        updatedItems = list(map(update_content_item_cover_image, contentItems))
+
+        #Sets all content items without a cover image to their parent's cover image
+        pg_hook.run(
```
vinnyjth, line 90: sweet!

dags/rock_cover_image.py
```diff
@@ -42,51 +49,24 @@ def fetch_and_save_cover_image(ds, *args, **kwargs):
...
+            True, 
+            (coverImageId, contentItemId))
+
     def map_content_items(content_item):
 
         imageAttributes = get_images(content_item)
-        coverImageId = get_best_image_id(imageAttributes)
+        coverImageId = get_best_image_id(imageAttributes, pg_hook)
 
         if(coverImageId):
```
vinnyjth, line 65: one more scenario I thought of :) if the `coverImageId` becomes `None`, we should handle that and `unset` the `coverImageId` (if an admin deletes an image off a content item, for example)

vinnyjth: I can make a separate note for this in basecamp, we can keep this PR focused.

vinnyjth (changes requested):
Getting close! Nice work @tomthornton

dags/rock_cover_image.py
vinnyjth: @tomthornton Take a look at this comment, I think we still need to switch over to the compound `Content`/`Attribute` ID

tomthornton:
@vinnyjth updated the ID ^^ This should be good to go.

tomthornton:
@vinnyjth good catch ^^ still getting used to this new id format for media

tomthornton (commented):
Got it. Content Item Id is first

vinnyjth:
> Gotcha. Does this mean we will need to update the media shovel to reflect this?

it was actually, but it looks like we accidentally overwrote the change in a recent PR. That's my bad for not catching it. I reverted it as a commit in this PR. 

https://github.com/ApollosProject/apollos-shovel/pull/10/commits/47b92a48251f483083de5e46d26534e0df5edf01

dags/rock_cover_image.py
```diff
@@ -38,55 +25,44 @@ def fetch_and_save_cover_image(ds, *args, **kwargs):
...
 
     def map_content_items(content_item):
+        def get_best_image_id (images):
+            imageId = None;
+            if(len(images) > 1):
+                squareImages = list(filter(lambda attribute: 'square' in attribute['Key'].lower(), images ))
+                if(len(squareImages) > 0):
+                    imageId = squareImages[0]['Id']
+                else:
+                    return images[0]['Id']
```
vinnyjth, line 44: I think this early return here is causing problems. When running the dag locally I'm getting conflicting `IDs` since this ID isn't being turned into a media ID.

dags/rock_cover_image.py
```diff
@@ -38,55 +25,44 @@ def fetch_and_save_cover_image(ds, *args, **kwargs):
...
+            elif(len(images) == 1):
+                imageId = images[0]['Id']
+
+            if(imageId):
+                concatImageId = str(content_item['Id'])  + '/' + str(imageId)
+                return pg_hook.get_first('SELECT id FROM "media" WHERE "originId" = %s', (concatImageId))[0]
```
vinnyjth, line 50: Related to my comment above, what if we returned the `originId` here and kept the above code?

vinnyjth (commented):
I tested the PR locally and ran into a few issues, but they should be easy fixes! I think we just need to settle on `get_best_image_id` returning a media.id or a `media.originId`, and reacting accordingly in `update_content_item_cover_image`

tomthornton:
@vinnyjth` I think the problem was that `return pg_hook.get_first('SELECT id FROM "media" WHERE "originId" = %s', (concatImageId,))[0]` was missing that comma after `concatImageId`. This is working for me now.

vinnyjth:
> @vinnyjth I think the problem was that `return pg_hook.get_first('SELECT id FROM "media" WHERE "originId" = %s', (concatImageId,))[0]` was missing that comma after `concatImageId`. This is working for me now.

My bad for removing that, are you ready for me to review this again?

tomthornton:
@vinnyjth yep!

vinnyjth:
@tomthornton Something still must be off :/ 

Here are the steps I took, and here is what I'm seeing.

1. Truncated my media table with cascade locally
2. Truncated my contentItems table with cascade locally. 
3. Re-ran the backfill dag and confirmed that there were no errors. 

Here's what I'm seeing in my contentItems table. 
image

and in my media table

image

From what I can tell, looking at the code, I believe you're `concating` the wrong media image id in the `update_content_item_cover_image` method. Looking at the logs of the `dag` I'm seeing what looks like a combination of a media `postgres` ID and a media `originId`

```
[2021-07-07 18:11:44,481] {dbapi.py:180} INFO - Running statement: UPDATE "contentItems" SET "coverImageId" = (SELECT id FROM "media" WHERE "originId" = %s) WHERE "originId"::Integer = %s, parameters: ('21/e5994c27-2de9-4261-81f3-e02c65f5674f', 21)
[2021-07-07 18:11:44,484] {dbapi.py:186} INFO - Rows affected: 1
```

I think there's two things you can do to resolve this - 

1. Stop querying for the `postgresId` in the `get_best_cover_image` method and pass the originId to the `update_content_item_cover_image`, eliminating the need for the pg query inside `get_best_cover_image`

2. You are passing the `media.id` to the `update_content_item_cover_image` method, so you should be able to eliminate the subquery instead of your sql and use the `media.id` directly.

tomthornton:
@vinnyjth that's my bad. I had tested it yesterday and it was working... but I don't know why it did. I ran this and it worked for most of the content items but it didn't update based on the parent content item with that SQL query you wrote.

dags/rock_cover_image.py
```diff
@@ -38,55 +25,46 @@ def fetch_and_save_cover_image(ds, *args, **kwargs):
...
+                if(len(squareImages) > 0):
+                    imageId = squareImages[0]['Id']
+                else:
+                    imageId = images[0]['Id']
+            elif(len(images) == 1):
+                imageId = images[0]['Id']
+
+            if(imageId):
+                concatImageId = str(content_item['Id'])  + '/' + str(imageId)
+                return pg_hook.get_first('SELECT id FROM "media" WHERE "originId" = %s', (concatImageId,))[0]
```
vinnyjth, line 52: is there any reason you decided to keep this query?

vinnyjth (approved):
Looking good! Nice work landing the plane

dags/rock_cover_image.py
tomthornton: Don't we still need the postgres media id? I guess we could do this in the update query... I may shift it over once we get to the next part which is setting the cover image as null if it's deleted

vinnyjth: It's either fetch the media id here, or use the `originId` in the  `update_content_item_cover_image` via a subquery. A bit more efficient since it takes us down to one query per insert, but not a huge deal :) Was curious if you had spotted something I missed.

PR #17, opened 2021-07-02
feat: Content item Typename support

vinnyjth:
Right now, matches what you would put into the config.yml 1-1


dags/rock_media.py
```diff
@@ -4,6 +4,23 @@
 
 import requests
 
+
+def is_media_image(attribute, contentItem):
```
vinnyjth, line 8: Had to move these to the root so I could import/export them

vinnyjth:
image

tomthornton (approved):
Interesting. What is the reason we are putting ApollosId and ApollosType?

vinnyjth:
@tomthornton It is indeed a little superfluous.

The type is important to have by itself because it's something you might filter by, and the `apollosId` is, in our worlds, always a concatenation of the `id` and the `apollosType`. However, there are situations where we might make the apollosId something different (like the rock id + the apollos type) so it's important to keep them separate.

PR #19, opened 2021-07-06
fix: Update Rock Image URL to be dynamic

tomthornton:


dags/rock_media.py
```diff
@@ -61,7 +61,8 @@ def get_media_value( attribute  ):
             attributeValue = contentItem['AttributeValues'][attributeKey]['Value']
 
             if(mediaType == 'IMAGE'):
-                return "https://rock.apollos.app/GetImage.ashx?guid=" + attributeValue if len(attributeValue) > 0 else ''
+                rock_host = (Variable.get(kwargs['client'] + '_rock_api')).split("/api")[0]
```
vinnyjth, line 64: Nice work!

PR #20, opened 2021-07-07
feat: Feature Operator

vinnyjth:
This PR mimics the `ContentItem.getFeatures` code from Core. I didn't port in the whole method, since some of it we were planning on deprecating :)


vinnyjth:
Screen Shot 2021-07-07 at 12 19 54 PM

dags/rock-content-item-backfill-core-dag.py
```diff
@@ -28,53 +29,59 @@
          # catchup=False # enable if you don't want historical dag runs to run
          ) as dag:
 
-    t0 = PythonOperator(
+    base_items = PythonOperator(
```
vinnyjth, line 32: I renamed the operators with actually helpful names, they were getting hard to manage :)

dags/rock_features.py
```diff
@@ -0,0 +1,198 @@
...
+        pg_hook.insert_rows(
+            '"features"',
+            list(insert_features),
+            columns,
+            0,
+            True,
+            replace_index=('"parentId"', '"type"', '"data"')
+        )
+
+        add_apollos_ids = """
```
tomthornton, line 189: It appears that "originType" is not a column on the features table. Are there more recent migrations that adds this?

tomthornton (changes requested):
Getting all of the features imported but the apollos ID isn't updating because there doesn't seem to be an `originType` column on the features table

dags/rock_features.py
vinnyjth: Good catch! That's a miss on my part.

vinnyjth:
@tomthornton Updated to accommodate that this table doesn't have/need an originId.

tomthornton (approved):
Great looks good!

PR #25, opened 2021-07-12
Fix multiple attribute values not being shoveled

vinnyjth:
For the attribute values that contain multiple personas, they show in rock like this.

`guid,guid,guid`. 

This caused issues shoveling, we need to split them up into attribute values.


dags/rock_tags.py
```diff
@@ -226,6 +228,13 @@ def attach_persona_tags_to_content(ds, *args, **kwargs):
         headers=headers
     ).json()
 
+    def split_to_multiple_attributes(attribute):
+        attributes = attribute["Value"].split(',')
+        return map(lambda a: {'Value': a, 'EntityId': attribute["EntityId"]}, attributes)
+
+    split_attribute_values = map(split_to_multiple_attributes, dataview_attribute_values)
+    attribute_values = reduce(lambda x,y: x + y, split_attribute_values)
```
tomthornton, line 236: Nice! Good Idea. I've been missing using `reduce()`

tomthornton (approved):

PR #26, opened 2021-07-12
feat: Content item active status

tomthornton:


dags/rock_content_items.py
```diff
@@ -143,7 +163,7 @@ def fetch_and_save_content_items(ds, *args, **kwargs):
         print(params)
 
         r = requests.get(
-                f"{Variable.get(kwargs['client'] + '_rock_api')}/ContentChannelItems",
+                f"{Variable.get(kwargs['client'] + '_rock_api')}/ContentChannelItems?$expand=ContentChannel",
```
vinnyjth, line 166: You can probably move this `$expand` to line 153 :) will make it easier to find / maintain in the future

tomthornton: Yep you're right. Thanks!

vinnyjth (approved):

vinnyjth:
Ran great! Code looks good. Thanks @tomthornton!

PR #27, opened 2021-07-12
feat: Rock content item delta dag

tomthornton:


vinnyjth:
@tomthornton were you able to test this to make sure it was working correctly?

tomthornton:
@vinnyjth Yeah I tested it and it worked, but when I came back to it, it had crashed my webserver docker instance. I restarted, ran again, and it worked, but crashed it again. Trying to find out what's going on.

vinnyjth:
@tomthornton are you still working on this? The branch is in an odd state, I don't see any new work 🤔

tomthornton:
@vinnyjth fixed a filter bug for the tags. Everything should be running as expected now.

vinnyjth:
Awesome! were you able to test some of the backfill situations? 

Here are a few things we should expect to happen 

- Adding a new cover image should work 
- Adding a new item should cause that item to show up in postgres
- Giving an item a parent of another item should set that item's parentId
- Creating a new persona dataview should add a new tag in the db
  - and people should get that new tag in the db 
- Adding a new feature field on an item should create a new feature in the db

vinnyjth:
It might make it easier if you set the interval to 5/10 minutes instead of 30 for the sake of testing

vinnyjth (changes requested):
Looking good @tomthornton - thanks for putting all the work into testing this! 

I'm a bit worried that it's going be a bit easy to de-sync the full sync and the delta sync. How do you feel about combining the two files, and then creating the `dags` dynamically, once with the backfull variable set and once without? 

I think you'll need to define a function that returns a dag, and then call that function twice. The `client` and `do_backfill` could become the arguments of the dag creating function :) 

https://www.astronomer.io/guides/dynamically-generating-dags

tomthornton:
@vinnyjth yeah, I think that's a good idea. I was thinking the same thing. I'll try it out!

tomthornton:
@vinnyjth just updated with the dynamic dag, that was a great idea. We could take this a step further and dynamically do it for each church.

vinnyjth:
> @vinnyjth just updated with the dynamic dag, that was a great idea. We could take this a step further and dynamically do it for each church.

Love it! 

I'm all for it, but we need to make sure we can be granular on a church by church basis. As in, Newspring uses the people shovel, but not the ContentItem shovel. If you have a good route in mind for implementing that, go for it!

tomthornton:
@vinnyjth granular just in terms of what shovels they use? That shouldn't be too hard to configure.

Is this good to go ahead and merge?

dags/rock-content-item-core-dag.py
```diff
@@ -0,0 +1,122 @@
...
+    'depends_on_past': False,
+    'email_on_failure': False,
+    'email_on_retry': False,
+    'retries': 1,
+    'retry_delay': timedelta(minutes=5),
+}
+
+def create_content_item_dag(dag_name, schedule_interval, do_backfill):
+    dag = DAG(dag_name,
+            start_date=datetime(2021, 6, 21),
```
vinnyjth, line 25: Let's make the start time configurable as well :) this is something every client will need to set.

vinnyjth (approved):

vinnyjth:
I left you one comment, we'll need to be able to change the start time easily. We can do that in this PR or the next though!

tomthornton:
@vinnyjth as a variable in webflow or in the code?

vinnyjth:
> @vinnyjth as a variable in airflow or in the code?
@tomthornton I would make it configurable in the same way you `made`, `schedule_interval` `do_backfill`  configurable :)

PR #28, opened 2021-07-15
chore: Add order to content item connections

tomthornton:


dags/rock_content_items_connections.py
```diff
@@ -86,13 +86,14 @@ def update_content(obj):
                 'ContentItemsConnection',
                 get_postgres_id(obj['ChildContentChannelItemId']),
                 get_postgres_id(obj['ContentChannelItemId']),
+                obj['Order']
```
vinnyjth, line 89: Can you add a reference to this field to the comment on line 79? I've been trying to do that to make code easier to parse.

dags/rock_content_items_connections.py
```diff
@@ -49,7 +49,7 @@ def get_postgres_id(id):
             "$top": top,
             "$skip": skip,
             # "$expand": "Photo",
-            "$select": "Id,ChildContentChannelItemId,ContentChannelItemId",
+            "$select": "Id,ChildContentChannelItemId,ContentChannelItemId,Order",
```
vinnyjth, line 52: 👍

vinnyjth (changes requested):
Looking good! One teeny tiny tweak :)

dags/rock_content_items_connections.py
tomthornton: @vinnyjth yep!

vinnyjth (approved):

PR #31, opened 2021-07-19
feat: Remove ContentItemConnections when child content item is removed

tomthornton:


tomthornton:
@vinnyjth I've rebased on master to include the deleted content item dag so that this one can depend on it. That way it takes care of parents being deleted and children being removed.

vinnyjth:
@tomthornton Correct me if I'm wrong, but I think you've already solved this :D 

image

Since the contentItemCategories table has the `ON DELTE CASCADE` constraint, it appears to delete the join table correctly when the appropriate content item is created. Same thing with the `parentId` :)

vinnyjth:
@tomthornton I tested this on `master` and it appears to work, deleting an item with children deleted the join table row, as well as nulling out the `parentId` field on the child.

tomthornton:
@vinnyjth you're right I did! I think I wrote this before I realized you could do it with foreign key constraints. We can delete this branch.

PR #32, opened 2021-07-20
Rename ButtonLinkFeature ButtonFeature

vinnyjth:
I realized the typename in core is actually ButtonFeature - woops!


tomthornton (approved):
Great :)

PR #34, opened 2021-07-21
chore: Update SQL naming conventions

tomthornton:
Need to have this PR's migrations to work

https://github.com/ApollosProject/apollos-apps/pull/2075


vinnyjth:
@tomthornton the core PR is merged, so we can move forward on this :) 

https://github.com/ApollosProject/apollos-apps/pull/2080

One ask though - can we add support for using the old column names in the `rock_people` and `rock_campus` names? It will make it easier to update Newspring gradually. I'm happy to do the work for that if you've got a lot on your plate.

dags/rock_campus.py
```diff
@@ -15,61 +16,102 @@ def safeget(dct, *keys):
...
-        SELECT id from campuses
-        WHERE "originType" = 'rock' and "apollosId" IS NULL
-        """
-
-        for new_id in pg_hook.get_records(users_without_apollos_id_select):
-            apollos_id_update = """
-            UPDATE campuses
-            SET "apollosId" = %s
-            WHERE id = %s::uuid
+        if(kwargs['use_underscore']):
```
vinnyjth, line 42: Let's make sure and `safeget` this value, so we don't throw an error if it's not set :) We should default to `True` if it's unset.

tomthornton:
@vinnyjth changed the format to make it easier. We can just delete the file later.

vinnyjth (approved):

PR #36, opened 2021-08-02
chore: Organizes files and creates dags dynamically

tomthornton:
@vinnyjth I figured it might be a good idea to clean up the shovel a little bit and create the dags dynamically. Let me know what you think


vinnyjth:
Duplicated this work in master and merged. Thanks for all the hard work here!

PR #40, opened 2021-08-04
Move personas/tags to their own DAG

vinnyjth:
They take much longer than the rest of the other content item item operations, and don't have a strict dependance, so I think it makes sense to split them off. They can also run at a less frequent interval without issue.


tomthornton (commented):
Good idea. I think that makes sense for them to run less often.

tomthornton (approved):

PR #42, opened 2021-08-05
Foolproof dag creation

vinnyjth:
This takes the guesswork out of naming the `dag` correctly :) I was having trouble getting these correct consistently, so now creating a dag also returns the name of the dag.


vinnyjth:
Note - this will cause the `content` dags to be renamed, creating duplicates. This should be okay since none of the content dags are in production (and the new dag will continue to run fine). The duplicates will have to be manually removed and won't run.

tomthornton (approved):
Thank you! I thought the same thing. Naming these dags is tough with all of the descriptors. I'm glad we have a consistent way of doing it now.

PR #43, opened 2021-08-05
Convert content item shovel task to a class

vinnyjth:


vinnyjth:
Posted originally on slack - but maybe this belongs here instead :) 

@tomthornton  - how do you feel about the direction of this pr?
I did three things
- Move all the functions into a ContentItem class
- Check to see if a user passed a custom class when the operator runs
- Instantiate the class (either default or custom) with the kwargs and call a named method on that class
- 
The idea here is that rather than doing what I was doing with media and making methods overridable by passing them down through kwargs, you would instead make a new ContentItem class that extends the original, overide your methods, and pass that to the runner. It’s a lot more powerful

vinnyjth:
@tomthornton this is all ready for review! You can see an example of how to do a method override in the `newspring_content_item_dag` now.

tomthornton (approved):
Looks good!

PR #44, opened 2021-08-06
Setup cover image operator to perform delta update

vinnyjth:


vinnyjth:
@tomthornton This is also ready for review now. It should dramatically speed up the `cover_image` task on larger apps when running the delta run.

tomthornton (approved):

PR #45, opened 2021-08-09
Gets river valley videos to show up

vinnyjth:


tomthornton (approved):
This makes sense. Thanks @redreceipt  @vinnyjth for doing this!

PR #46, opened 2021-08-09
fix: Update scriptures feature

tomthornton:
This should add the ability to pull in scriptures from River Valley


dags/rock/rock_features.py
```diff
@@ -61,9 +60,20 @@ def get_features(self, content):
         # Add features from a key value list
         # This is now the only way to add scripture and text features
         # We previously supported a textFeature and
+        
         key_value_features = self.parse_key_value_attribute(
             safeget_no_case(content, "AttributeValues", "Features", "Value") or ""
         )
+
+        attributeValues = safeget(content, "AttributeValues")
```
vinnyjth, line 68: Two things :D 

- I think you can avoid doing two `safeget`s if you combine the nested keys you are looking for. For example `scripture = safeget(content, 'AttributeValues', 'Scriptures')

- It might be worth using `safeget_no_case` here since that will also catch attribute values named `scripture`. Rock doesn't standardize the casing of attribute values unfortunately so user's can input whatever they want.

dags/rock/rock_features.py
```diff
@@ -185,7 +195,7 @@ def run_fetch_and_save_features(self):
                 # "$select": "Id,Content",
                 "loadAttributes": "expanded",
                 "$orderby": "ModifiedDateTime desc",
-                "attributeKeys": "features, comments, buttontext, buttonlink",
+                "attributeKeys": "features, comments, buttontext, buttonlink, Scriptures",
```
vinnyjth, line 198: Good catch! This always trips me up

vinnyjth:
@tomthornton I left you some comments, once I hear back from you I'll look the PR over again :)

tomthornton:
@vinnyjth updated! decided not to use safeget_no_case

vinnyjth (approved):
Looking good! Runs well, thanks!

PR #48, opened 2021-08-11
fix summaries

redreceipt:


dags/rock/rock_content_items.py
```diff
@@ -93,8 +89,8 @@ def map_content_to_columns(self, obj):
         )
 
     def create_summary(self, item):
-        summary_value = safeget(item, "AttributeValues", "Attribute", "Value")
-        if summary_value and summary_value is not "":
+        summary_value = safeget(item, "AttributeValues", "Summary", "Value")
```
redreceipt, line 92: this is the important change, the rest is my linter

redreceipt: i'm gonna see about putting everything else back so this isn't confusing

vinnyjth (approved):

PR #49, opened 2021-08-11
chore: Add channel cover image dag

tomthornton:


redreceipt:
@tomthornton @vinnyjth this seems overkill to me, we just want the channel image to be the fallback content item in the event there is no other. seems like that could be done entirely inside the cover image class https://github.com/ApollosProject/apollos-shovel/blob/master/dags/rock/rock_cover_image.py

we don't necessarily care about rock channels other than mapping them to categories. take my opinion with a grain of salt though, I don't know any DB schema arch strategies you guys have built thus far.

dags/rock/rock_channel_image.py
```diff
@@ -0,0 +1,120 @@
...
+        channels = requests.get(
+            f"{Variable.get(self.kwargs['client'] + '_rock_api')}/ContentChannels",
+            params=params,
+            headers=self.headers,
+        ).json()
+
+        channelsWithImages = map(
+            self.map_channels,
+            filter(
+                lambda channel: safeget(channel, "AttributeValues", "BlogImage"),
```
redreceipt, line 75: `BlogImage` is naive for all churches, we should use the regular content image utility functions to find the images -  "if `image` in the attribute name" etc...

redreceipt:
...all that being said, if we deem this is correct code and could be simply be improved, could we run the shovel against their production instance using this PR just to get images in the DB, and then we can talk about finishing the PR? I just wanna get their app looking good since they wanna launch Sunday.

dags/rock/rock_channel_image.py
vinnyjth: 👍 I agree with @redreceipt here, we should follow the same logic we use to search a content item's attributes for images. 

While this code is in a separate task, you could potentially co-locate the code within the `Media` class to get access to all its helper methods. Look at how the `tags` tasks are arranged for examples.

dags/rock/rock_channel_image.py
```diff
@@ -0,0 +1,120 @@
...
+                channels,
+            ),
+        )
+
+        channelsWithImageValues = filter(lambda channel: channel, channelsWithImages)
```
vinnyjth, line 80: 💥 

can this be further simplified to `filter(channelsWithImages)`?

dags/rock/rock_channel_image.py
```diff
@@ -0,0 +1,120 @@
...
+        else:
+            return None
+
+    def run_fetch_and_save_channel_image(self):
+        params = {
+            "loadAttributes": "expanded",
+            "$orderby": "ModifiedDateTime desc",
+        }
+
+        if not self.kwargs["do_backfill"]:
```
vinnyjth, line 61: Nice catch here!

dags/rock/rock_channel_image.py
```diff
@@ -0,0 +1,120 @@
...
+            keepalives=1,
+            keepalives_idle=30,
+            keepalives_interval=10,
+            keepalives_count=5,
+        )
+
+    def map_channels(self, channel):
+        print(channel["AttributeValues"])
+        channel_id = channel["Id"]
+        node_id = self.pg_hook.get_first(
```
vinnyjth, line 27: Do you think we could move this inside the `if image_value` block so we don't make a postgres query if we don't have to?

dags/rock/rock_cover_image.py
```diff
@@ -58,6 +58,19 @@ def get_best_image_id(self, images, content_item):
 
         return None
 
+    def get_channel_image(self, content_item):
+        rock_channel_id = content_item["ContentChannelId"]
+        node_id = self.pg_hook.get_first(
```
vinnyjth, line 63: Do you think we could combine this query, and the below query into a single query with a join? Would save on DB bandwidth :)

dags/rock/rock_cover_image.py
```diff
@@ -75,6 +88,15 @@ def map_content_items(self, content_item):
             self.update_content_item_cover_image(
                 {"ContentItemId": content_item["Id"], "CoverImageId": cover_image_id}
             )
+        else:
+            channel_image_id = self.get_channel_image(content_item)
+            if channel_image_id:
+                self.update_content_item_cover_image(
```
vinnyjth, line 94: 💯

dags/rock/rock_cover_image.py
tomthornton: Yep I was just trying to get it done yesterday. That's the query I reached out about. Do you mind writing it?

vinnyjth:
@redreceipt 
> this seems overkill to me, we just want the channel image to be the fallback content item image in the event there is no other. 

> we don't necessarily care about rock channels other than mapping them to categories. take my opinion with a grain of salt though, I don't know any DB schema arch strategies you guys have built thus far

The issue is that it's not a simple as "here's a url, it belongs on an image". We need to track that url on a media record, and that media record has to be attached to something. Otherwise we run into issues if that image ever changes, or if it's removed. 

That said, if you need to run this locally on River Valley to get those media files built out so you can move forward with the client, I full support that :) You could also "gasp" add those values by hand for now (assuming there's a limited number of areas that need those images)

vinnyjth (changes requested):
Left a bit of feedback in the PR. Looking really good! Just a few places to cleanup

dags/rivervalley-rock-content-item-dag.py
```diff
@@ -217,6 +222,9 @@ def create_rock_content_item_dag(church, start_date, schedule_interval, do_backf
 
         base_items >> [connections, deleted_content_items, add_categories]
 
+        channel_image.set_downstream(set_cover_image)
```
vinnyjth, line 225: Logic here looks great! What do you think about using the bitwise operators here? Instead of set_downstream? I noticed that the astronomer docs advices against mixing and matching the two syntaxes 

https://www.astronomer.io/guides/managing-dependencies

dags/rock/rock_channel_image.py
vinnyjth: @tomthornton Wanted to make sure you saw this comment :)

dags/rock/rock_channel_image.py
vinnyjth: @tomthornton I think we still need to think about this before merging, or make this PR specific to rivervalley.

vinnyjth (changes requested):
Hey @tomthornton! Looking really close, thanks for fixing the linter errors. I think we've bought ourselves enough time to get this right for all of the churches, (instead of just river valley) so we can take the time here to iron out some more details before we merge this 👍

dags/rock/rock_channel_image.py
vinnyjth: Looks like the PR review tool doesn't show context well - so I'll elaborate. We should use the same logic we use to determine what a cover image is on a content item to determine what the image is for a content channel :) looking for just the "BlogImage" attribute is too narrow and won't work for most churches.

tomthornton:
@vinnyjth made those changes :) this should make things run more smooth and consistent

vinnyjth (approved):

PR #52, opened 2021-08-14
adds algolia indexer to rivervalley

redreceipt:
if we like, this can be easily moved to its own church agnostic module


redreceipt:
actually images aren't quite right yet...

redreceipt:
now they are

vinnyjth (approved):

vinnyjth:
So excited for this! Let's get it merged and then bring it live for some other churches.

PR #53, opened 2021-08-16
feat: Simplify status field

vinnyjth:
We don't want to take into account the publishAt / expiresAt fields when considering if something is active. We'll let the API take care of that.


tomthornton (approved):
Looks good

vinnyjth:
Don't merge until RiverValley is updated, otherwise content that should be live in the app will stop being live in the app.

tomthornton:
@vinnyjth RV updated and deployed. Good to merge?

PR #54, opened 2021-08-17
fix: Add feature priority

tomthornton:


dags/rock/rock_features.py
```diff
@@ -168,6 +168,14 @@ def get_features(self, content):
 
         return features
 
+    def get_feature_priority(self, feature_type):
+        if feature_type == "AddComment":
```
vinnyjth, line 172: any reason you went with a `descending` order? what will we do if we want something to be featured before `AddComment`? 

An example of this would be scripture - one bug that I believe is likely happening but we're not seeing the full ramifications of, is that the `scriptures` field likely has an order that churches would like the scripture to show up in. We should consider using this `priority` field to ensure that the scriptures show in the feature list in the same order they were listed in the string we are parsing.

tomthornton: I was thinking this was a one-off scenario, but you're right. I can get the ordered `scriptures` and use that to calculate the `priority` of `AddComment` and `CommentList`. Are there any other scenarios for features we should be aware of?

vinnyjth: I can't think of any off the top of my head. In general, the features we get from `keyValueList` are in the correct order, so building an array of features from that list should set priorities according to the order in htat list.

vinnyjth: It might be worth consulting how the features are pushed to the features array in the Rock ContentItem.getFeatures method - that represents the exact order features should be presented in.

tomthornton: Updated the feature list to use the index of the array. Also reordered button link because it was at the bottom in the rock datasource

vinnyjth (approved):
Looks good! I like the new approach. We'll need to wait until the `core` PR is merged before we can merge this.

vinnyjth:
@tomthornton If you fix the linter on this, we can merge it.

PR #55, opened 2021-08-24
Fix backfill dates

vinnyjth:
The old delta dates were hardcoded to the start of the day. Items were being missed. I updated them so they are now accurate to the second, and also take into account the Rock timezone. We'll have to add support for additional timezones moving forward, likely as a Config option or something similar.


dags/utilities.py
```diff
@@ -21,3 +24,12 @@ def safeget_no_case(dct, *keys):
...
             return None
     return dct
+
+
+def get_delta_offset(kwargs):
+    local_zone = pytz.timezone("EST")
+    execution_date_string = (
+        kwargs["execution_date"].astimezone(local_zone).strftime("%Y-%m-%dT%H:%M:%S")
+    )
+    print(execution_date_string)
```
vinnyjth, line 34:
```suggestion
```

tomthornton (approved):
Oh so that's why items were being missed. Makes sense. Great job!

PR #57, opened 2021-09-03
Pick up updated items by their attributes when shoveling

vinnyjth:
Steps to test! 

1. Run shovel locally pointed at the core rock instances
2. In the core rock, find and update a content item's attribute.
    - Something like `Summary` or `Image` or a feature. 
3. Rerun the most recent `core_content_item_dag` by clearing it. Not the backfill. 
4. You should see the updates show up in your database. 

Caveats: 
This will only work as long as less than 16 item's attributes are being updated at any given point in time. (https://www.rockrms.com/ask/developing/2169)


vinnyjth:
image

dags/utilities.py
```diff
@@ -28,6 +29,34 @@ def safeget_no_case(dct, *keys):
...
+        params={
+            "$filter": f"({get_delta_offset(kwargs)}) and Attribute/EntityTypeId eq {item_entity_id}"
+        },
+        headers={"Authorization-Token": Variable.get(kwargs["client"] + "_rock_token")},
+    ).json()
+
+    entity_ids = set([a["EntityId"] for a in entity_id_result])
+
+    # We can't fetch more than 16 items at once due to a rock limitation.
+    # These items will need to be manually touched, or grabbed during a periodic backfill.
```
conrad-vanl, line 52: 🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮

vinnyjth: I know 😭 😭 😭 😭 😭 😭 😭 😭 😭

conrad-vanl: This comes into play when there are more than .... 16 custom attributes right?

vinnyjth: I can talk to you more about why I believe this is acceptable if you wish.

vinnyjth: This comes into play if more than 16 times have their custom attributes edited at once (in a 5 minute period)

conrad-vanl: Is there a way you could trigger a backfill to be scheduled in this code branch? how often do backfills run?

vinnyjth: And if that happens, we'll still nab them if their items content items themselves are touched. Maybe instead of bailing out of the whole filter, we could omit everything after the first 16 items?

conrad-vanl: my worry with the silent fail is that this goes unnoticed until we get a support ticket coming in asking why items aren't updating

vinnyjth: > Is there a way you could trigger a backfill to be scheduled in this code branch?
We could but I'm not sure if that's the right solve. 

> my worry with the silent fail is that this goes unnoticed until we get a support ticket coming in asking why items aren't updating. 
Yah I agree. I was trying to decide if an overt failure (which would cause no items to be updated) was worse than a silent failure (where we still might capture the changes anyway)

vinnyjth: What do you think about some sort of logging to capture when the above situation is hit? One thing I'm not sure about is how often this will happen to begin with.

conrad-vanl: > We could but I'm not sure if that's the right solve.
Why?

The other solution is what....write another function that grabs them 16 at a time? like a Promise.all() situation in Javasciprt

tomthornton: I've got my features fetching like a backfill each time. It still runs pretty fast and only updates/deletes/adds exactly what's needed.

vinnyjth: It's going to happen at some point or another - thank's Murphy - but by keeping the delta timing low (5 min) I imagine it will be pretty infrequent/never on most churches.

conrad-vanl: While it is an edge case, it's an edge case that I think is bound to happen, and we need some solve for it...queuing a backfill to start in the next 30 minutes or something doesn't seem like a bad solve to me?

vinnyjth: > The other solution is what....write another function that grabs them 16 at a time? like a Promise.all() situation in Javascript

This often slams rock pretty hard on larger churches, and is difficult to integrate into our shovel (but we could do it)

vinnyjth: > Why?
Mostly because Backfills can take a long time and tend to hit rock hard. imo it's something that needs to be done with some amount of finesse.

conrad-vanl: How often do we have backfills run?

vinnyjth: Here's what I'm at in this PR 

1. I wanted to stop the bleeding (frequent support request about missing data) 
2. I wanted to integrate essentially seamlessly into our existing querying strategy 
3. I wanted it to scale at least decently across all our clients 

Here's what I didn't set out to do in this PR 

1. Solve the problem permanently :)  
2. Rewrite how we fetch and paginate across updated items across all of our dags that fetch rock content.

vinnyjth: @conrad-vanl Right now the intention is to run them at these times

1. When we boot up the church for the first time 
2. If we ever add new attributes to our schema that old items need

vinnyjth: So I would rather be notified that we need to run a backfill, and then go run it, then have it happen automatically.

conrad-vanl: Where I'm at:

1. If we have more then 16 items updated in a 5 minute period, there's likely some automated process going on and there could be a TON of items updated.... running a backfill might be the best solution
2. Silently failing seems really dicey. If we throw an error....that seems just as bad since nothing gets shoveled

So if we can have some way to be notified that a backfill is needed...I'm game with that. 


This is probably not something to do now - but it almost seems like we could have a backfill setup to run nightly, but have that backfill only do something if there has been _something_ in the last 24 hours that said it should run....almost like a data integrity tracker. If there's shovel tasks that are erroring, if there are misses (like this one), etc. Almost as an automated way to "clean up" a potentially bad state. 🤷‍♂️

vinnyjth: > This is probably not something to do now - but it almost seems like we could have a backfill setup to run nightly, but have that backfill only do something if there has been something in the last 24 hours that said it should run....almost like a data integrity tracker. If there's shovel tasks that are erroring, if there are misses (like this one), etc. Almost as an automated way to "clean up" a potentially bad state. 🤷‍♂️

I've thought about this as well, but in a different context (more around the DB cleanup / catching new attributes). I'm game for that, and I think the biggest thing holding me back from wanting automated backfill runs is that the runs in the past have actually failed in a few places and need to be babysat to finish. I think I've cleaned up a lot of the situations that led to those failures (most of them query/rock optimizations) but I want to prove to ourselves we can run the backfill repeatedly without failures first :)  

I also think that this 
> Rewrite how we fetch and paginate across updated items across all of our dags that fetch rock content.

Is a valid option going forward. Something along the lines of what you said in slack about shoveling based on fields/attributes as much as possible. The issue with doing that now is that 

1. It's super time consuming.
2. We have no tests, and no way to write tests. Making changes right now can be really scary. I'm going to reach out to Astronomer for ideas on this next week.

vinnyjth: Also @conrad-vanl we have a third option. We could add an Apollos Plugin endpoint that adds the functionality very easy. It's a line of .net code away.

conrad-vanl: You can give that task to Rich if you want...I'm just nervous that if there's more then 16 there could be a TON (such as an automated process) and that might overwhelm rock anyways. like I'm sure there's a good reason this is limited to just 16 ids 🤷‍♂️

vinnyjth: We wouldn't need to pass ids if we get Rich involved, we could do the entire query using the .net object model

conrad-vanl: Well either way lets get this merged then you're welcome to talk with him on it...I do agree in priority in getting this implemented to save on support volume

conrad-vanl (approved):

dags/utilities.py
vinnyjth: 👍 I will look into getting some sort of notification setup ASAP

PR #58, opened 2021-09-03
Update rock_features.py

tomthornton:
@vinnyjth here you can see in the Airflow log how many are updated, deleted, added. Sending you the test file via slack


vinnyjth:
@tomthornton I think this PR is a good candidate for unit tests. Do you think it would be worth adding tests over some cases, such as - 

- Adding a new features to the beginning of a list of features
- Deleting a feature from the middle of a list of features
- Changing an existing feature 

Using VCR, the data will be pulled from rock for all the requests the task makes. You'll then be able to go into the generated VCR files and edit them to match the desired test data. Or something else :)

dags/rock/rock_features.py
```diff
@@ -219,16 +286,79 @@ def map_feature_to_columns(self, obj):
...
+                """,
+                True,
+                (self.deleted_features,),
+            )
+
+    def update_features(self):
+        print(f"Updated Features: {len(self.updated_features)}")
+        if len(self.updated_features) > 0:
+            for feature in self.updated_features:
+                self.pg_hook.run(
```
vinnyjth, line 305: 👍

dags/rock/rock_features.py
```diff
@@ -219,16 +286,79 @@ def map_feature_to_columns(self, obj):
...
+                "updated_at",
+                "apollos_type",
+                "data",
+                "type",
+                "parent_id",
+                "parent_type",
+                "priority",
+            )
+
+            self.pg_hook.insert_rows(
```
vinnyjth, line 344: How do you feel about using the new `find_supported_fields` (https://github.com/ApollosProject/apollos-shovel/blob/master/dags/rock/utilities.py#L18) function? It should help detach the schema  from schema changes and gets rid of the ordering depending of insert_features/columns.

dags/rock/rock_features.py
```diff
@@ -219,16 +286,79 @@ def map_feature_to_columns(self, obj):
...
+                True,
+                replace_index=('"parent_id"', '"type"', '"data"'),
+            )
+
     def run_fetch_and_save_features(self):
         fetched_all = False
         skip = 0
         top = 10000
 
-        retry_count = 0
```
vinnyjth, line 227: Any reason you took out `retry_count` here?

dags/utilities.py
```diff
@@ -0,0 +1,42 @@
+import pytz
```
vinnyjth, line 1: I think this file is a duplicate? It was moved recently to `rock/utilities.py`

vinnyjth (commented):
Some quick feedback before I clone down the code itself and spin through it :)

dags/rock/rock_features.py
```diff
@@ -28,6 +28,35 @@ def __init__(self, kwargs):
         )
 
         self.rock_content_cache = {}
+        self.all_postgres_features = []
```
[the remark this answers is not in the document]
tomthornton: I figured that it would be easy to access these in scope throughout all of the functions

dags/rock/rock_features.py
```diff
@@ -28,6 +28,35 @@ def __init__(self, kwargs):
...
+        self.deleted_features = []
+        self.updated_features = []
+        self.added_features = []
+
+    def fetch_all_postgres_features(self):
+        allPostgresFeatures = self.pg_hook.get_records(
+            """
+                SELECT content_item.origin_id, array_to_json(array_agg(fo)) AS features
```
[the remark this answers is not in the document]
tomthornton: Yeah, sure. `ORDER BY` doesn't really do anything.

dags/rock/rock_features.py
tomthornton: Yep, great idea. I like the idea of not having to go back and change things like this when the schema changes.

dags/rock/rock_features.py
tomthornton: No reason, I think it got deleted by accident.

dags/utilities.py
tomthornton: Yep, it is. I'll delete it.

dags/rock/rock_features.py
```diff
@@ -88,6 +119,41 @@ def map_feature_priority(self, feature):
         obj = feature[1]
         return {**obj, "priority": index}
 
+    def check_postgres_features(self, rockContentItemFeatures, contentItemId):
```
vinnyjth, line 122: Can you use `snake_case` here as well please?

vinnyjth: I'll try and add the snake_case / camelCase sensitivity to the linter as well so the linter will warn you.

dags/rock/rock_features.py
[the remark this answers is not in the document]
vinnyjth: I think we might be using class attributes improperly here :) I'm not sure if we've ever articulated this, and I'm struggling to find where I've read about it and gotten my ideas from, but in general it's encouraged to track process specific variables as instance variables that are passed around between class functions. That's the pattern we adhere to in all of our other `Class`es, where instance attributes are always either connections to APIs are caches for storing data we don't want to fetch more than once. 

What would you think about refactoring with that in mind? I don't think it would be too much of a refactor, you would have to return the `updated/deleted/new` features from `get_features`. 

I think that will also be more idiomatic and easier to read. Right now it's hard to tell where and how these arrays are populated without tracing through the entire execution path of `get_features`. it also would give you something to return from `get_features` which would help eliminate the linter error that I think is popping up there

dags/rock/rock_features.py
tomthornton: Yep! I much prefer snake_case anyways but it's just a habit from JS

dags/rock/rock_features.py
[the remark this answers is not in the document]
tomthornton: Yeah I agree. Would it still be good to keep all of the postgres features as a class attribute?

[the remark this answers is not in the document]
tomthornton: Also, you're right about it being easier to read. I was just looking at the file and thinking the Feature class is huge and hard to read.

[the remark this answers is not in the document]
tomthornton: @vinnyjth refactored to remove the class attributes. Could probably do a better job with naming things but let me know what you think of the new approach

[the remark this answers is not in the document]
vinnyjth: Love it. It feels much easier to read.

vinnyjth:
> @tomthornton I think this PR is a good candidate for unit tests. Do you think it would be worth adding tests over some cases, such as -
> 
> * Adding a new features to the beginning of a list of features
> * Deleting a feature from the middle of a list of features
> * Changing an existing feature
> 
> Using VCR, the data will be pulled from rock for all the requests the task makes. You'll then be able to go into the generated VCR files and edit them to match the desired test data. Or something else :)

there's been a lot and back and forth on this PR, so I just wanted to make sure you saw my comment about tests. Tests on this task I think are in the "lot of work"  but "lot of payoff" category.

vinnyjth:
Code is looking good @tomthornton! I wasn't able to identify and issues when updating, adding, or deleting features in any combination. Once we have some tests we can merge this :)

tomthornton:
@vinnyjth yep been working on the tests! trying to figure out the best way to do these. I've been reading up on pytest and seeing what the best way to do this is.

tomthornton:
@vinnyjth added the tests!

vinnyjth:
Looking great! thanks @tomthornton

dags/newspring-rock-content-item-dag.py
```diff
@@ -91,7 +91,6 @@ def create_content_item_dag(dag_name, start_date, schedule_interval, do_backfill
         max_active_runs=1,
         schedule_interval=schedule_interval,
         default_args=default_args,
-        tags=["newspring", "content"],
```
vinnyjth, line 94: Hm, I don't think this should have been backed out.

dags/rock/rock_campus_media.py
```diff
@@ -55,11 +55,7 @@ def add_postgres_data_to_rock_media(self, campuses):
     def map_campuses_to_images(self, campuses):
         campuses_with_images = []
         for campus in campuses:
-            if (
```
vinnyjth, line 58: Same as above - I think this change needs to be reverted.

dags/rock/rock_media.py
```diff
@@ -15,7 +15,7 @@
 def getsizes(uri):
     # get file size *and* image size (None if not known)
     try:
-        file = urllib.request.urlopen(uri.replace(" ", "%20"))
```
vinnyjth, line 18: Same as above - I think this change needs to be reverted.

dags/newspring-rock-content-item-dag.py
tomthornton: Weird. I didn't edit this file. Must have happened in a merge with master

dags/rock/rock_content_items.py
```diff
@@ -165,6 +165,7 @@ def get_typename(self, item, config):
             )
             > 0
         )
+        print(is_media_item)
```
vinnyjth, line 168: Sorry @tomthornton, but these lines were added when they shouldn't have been. Probably related to that bad merge.

dags/rock/rock_content_items.py
```diff
@@ -183,7 +184,7 @@ def run_fetch_and_save_content_items(self):
 
         fetched_all = False
         skip = 0
-        top = 1000
+        top = 10000
```
vinnyjth, line 187: same here

dags/rock/rock_content_items.py
```diff
@@ -223,6 +224,7 @@ def run_fetch_and_save_content_items(self):
                     raise Exception(f"Rock Error: {rock_objects}")
 
                 retry_count += 1
+                skip += top
```
vinnyjth, line 227: and here

dags/rock/rock_content_items_connections.py
```diff
@@ -136,15 +136,14 @@ def run_set_content_item_parent_id(self):
                   content_item_connection.child_id AS id
            FROM
              (SELECT c.id,
-              count(cc.child_id) AS parents_count,
```
vinnyjth, line 139: I think changes to this file need to be reverted

dags/rock/rock_features.py
```diff
@@ -97,10 +173,8 @@ def get_features(self, content):
         # This is now the only way to add scripture and text features
         # We previously supported a textFeature and
 
-        key_value_features = list(
-            self.parse_key_value_attribute(
-                safeget_no_case(content, "AttributeValues", "Features", "Value") or ""
-            )
+        key_value_features = self.parse_key_value_attribute(
```
vinnyjth, line 176: this needs to be wrapped in `list()` iirc since we call `push` on it, and we also call `map` on it. This change was made in master recently.

dags/rock/rock_content_items.py
tomthornton: That's weird. I merged master, reverted it, and committed to this branch and it didn't say there were merge conflicts. Sorry about that.

vinnyjth (commented):
Really close - I found more lines that seem to be `reverts` of code in master. How are you pulling down `master`?

dags/rock/rock_features.py
tomthornton: Yep, I added it to the `parse_key_value_attributes` function

tomthornton:
@vinnyjth I resolved the merge conflicts in `rock_features.py` but when I ran the test again, I was getting an error for overwriting the content_item casettes. So I reverted the merge commit, tested again, and it worked. I think it was because I initially fixed the merge conflicts incorrectly.

dags/rock/rock_features.py
vinnyjth: oh nice!

vinnyjth (approved):

PR #60, opened 2021-09-07
Omit unsupported fields when inserting

vinnyjth:
image

This will prevent most most updates and additions to the shovel from breaking shoveling if the client app is not yet updated. We'll have to implement it across the board first, of course :) 

Best way to test is to run the core people/campus backfill shovel locally _after_ deleting the `active` column in your database (or any other column for that matter) the shovel should still run without failing


tomthornton (approved):
Nice! I think this makes sense. It's only going to happen more and more when new churches adopt the Shovel.

PR #62, opened 2021-09-08
Import Rock Prayers to Postgres

tomthornton:


dags/rock/rock_prayer_requests.py
```diff
@@ -0,0 +1,128 @@
...
+                'prayer_request',
+                prayer_requests,
+                columns,
+                0,
+                True,
+                replace_index=('"origin_id"', '"origin_type"'),
+            )
+
+            add_apollos_ids = """
+            UPDATE content_item
```
vinnyjth, line 110: This should probably be `prayer_request`

dags/rock/rock_prayer_requests.py
```diff
@@ -0,0 +1,128 @@
...
+            SET apollos_id = apollos_type || ':' || id::varchar
+            WHERE origin_type = 'rock' and apollos_id IS NULL
+            """
+
+            self.pg_hook.run(add_apollos_ids)
+
+            print(prayer_requests)
```
vinnyjth, line 117: You'll probably want to remove this :)

dags/rock/rock_prayer_requests.py
```diff
@@ -0,0 +1,128 @@
...
+            rock_objects = requests.get(
+                f"{Variable.get(self.kwargs['client'] + '_rock_api')}/PrayerRequests",
+                params=params,
+                headers=self.headers,
+            ).json()
+
+            if not isinstance(rock_objects, list):
+                print(rock_objects)
+                print("oh uh, we might have made a bad request")
+                print("top: {top}")
```
vinnyjth, line 80: I've spotted this in a few other places in the app (all my fault) ;) make sure you make these print statements formatted strings
```suggestion
                print(f"top: {top}")
```

dags/rock/rock_prayer_requests.py
```diff
@@ -0,0 +1,128 @@
...
+
+        while not fetched_all:
+            # Fetch prayer requests from Rock
+
+            params = {
+                "$top": top,
+                "$skip": skip,
+            }
+
+            """ if not self.kwargs["do_backfill"]:
```
vinnyjth, line 68: I asked this above, but I'll repeat it here. Any reason you're not setting up a `delta` dag?

vinnyjth (changes requested):
Hey @tomthornton - I haven't gotten a chance to review this in practice yet, but I left you some comments about the code after skimming it. I'll clone it down and give it a run a bit later. Thanks!

dags/core-rock-prayer-request-dag.py
```diff
@@ -0,0 +1,15 @@
+from airflow import DAG  # noqa: F401
+
+from datetime import datetime, timedelta
+
+from rock.rock_prayer_request_dag import create_prayer_request_dag
+
+# Default settings applied to all tasks
+
+start_date = datetime(2021, 6, 21)
+
```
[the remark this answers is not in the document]
tomthornton: I was thinking that Prayers into Postgres would only be a one-time backfill. Do Prayers in Rock come from places other than an Apollos app?

dags/rock/rock_prayer_requests.py
```diff
@@ -0,0 +1,128 @@
+from airflow.models import Variable
+from airflow.hooks.postgres_hook import PostgresHook
+from html_sanitizer import Sanitizer
+import nltk
```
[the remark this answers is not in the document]
tomthornton: Nope! I'll remove it.

dags/core-rock-prayer-request-dag.py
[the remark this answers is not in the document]
vinnyjth: They can (but normally don't). However right now moderation is only possible via Rock, so we'll need to keep whatever prayers we have in Rock updated in Postgres (and vice versa)

dags/rock/rock_prayer_requests.py
tomthornton: 👍

dags/rock/rock_prayer_requests.py
tomthornton: My reasoning was I thought since we're using strictly Postgres prayers we would only be migrating them over once

vinnyjth: Unfortunately no :/ We'll need to stay in sync with Rock until we have a mechanism to moderate without rock.

dags/core-rock-prayer-request-dag.py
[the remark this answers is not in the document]
tomthornton: Oh okay that makes sense.

dags/rock/rock_prayer_request_dag.py
```diff
@@ -0,0 +1,40 @@
...
+
+    with dag:
+        prayer_requests = PythonOperator(
+            task_id="fetch_and_save_prayer_requests",
+            # make sure you don't include the () of the function
+            python_callable=fetch_and_save_prayer_request,
+            op_kwargs={"client": church, "do_backfill": do_backfill},
+        )
+
+        prayer_requests
```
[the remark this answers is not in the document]
tomthornton: Yeah the linter doesn't like it. Do we just merge it anyways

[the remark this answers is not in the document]
vinnyjth: Nah, you should be able to remove it with not difference in functionality.

dags/rock/rock_prayer_requests.py
```diff
@@ -0,0 +1,128 @@
...
+                print("skip: {skip}")
+                skip += top
+                continue
+
+            skip += top
+            fetched_all = len(rock_objects) < top
+
+            prayer_requests = list(map(self.map_prayer_requests, rock_objects))
+
+            columns = (
```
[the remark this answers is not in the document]
vinnyjth: @tomthornton just making sure you saw this :)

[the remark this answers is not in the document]
tomthornton: @vinnyjth I actually didn't, thanks! I looked over this before but couldn't see any new comments. Looking into this now.

tomthornton:
@vinnyjth this should be ready :)

vinnyjth:
Thanks @tomthornton! Once you fix the linter errors I will take a look at it.

tomthornton:
@vinnyjth fixed the lint errors 👍

tomthornton:
@vinnyjth did we ever decide on what the fix for the lint error for `Klass` was?

vinnyjth:
> @vinnyjth did we ever decide on what the fix for the lint error for `Klass` was?

here is @redreceipt's recommendation which I support https://github.com/ApollosProject/apollos-shovel/pull/67#issuecomment-919423911

tomthornton:
Yep I agree. That works.

tomthornton:
@vinnyjth added the `approved` and `flag_count` fields from rock. Also updated the `created_at` time so that imported prayers from rock will always be when they were created, not when the dag runs.

tomthornton:
Also added the ability to delete prayer requests when deleted from rock. I figure we can extend this in the future to delete old/outdated prayer requests

dags/rock/rock_prayer_request_dag.py
```diff
@@ -0,0 +1,56 @@
...
+    "depends_on_past": False,
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 1,
+    "retry_delay": timedelta(minutes=5),
+}
+
+
+def create_prayer_request_dag(church, start_date, schedule_interval, do_backfill):
+    tags = [church, "prayer_request", "backfill"]
```
vinnyjth, line 22: I'm not quite sure what your goal is with the names / tags variables here :)

dags/rock/rock_prayer_requests.py
```diff
@@ -0,0 +1,184 @@
...
+
+            self.pg_hook.run(add_apollos_ids)
+
+    def run_delete_prayer_requests(self):
+        fetched_all = False
+        skip = 0
+        top = 10000
+        rock_prayer_requests = []
+        postgres_prayer_requests = self.pg_hook.get_records(
+            "SELECT origin_id, id FROM prayer_request"
```
vinnyjth, line 112: I found a subtle bug, that _might_ not ever be a real issue in production - but we should probally protect against. 

When I ran the delete task I had a single prayer _not_ attached to a rock prayer in my DB. After the task ran, that prayer was gone (deleted). What do you think about not touching (deleting) prayers that didn't come from rock? that might be the case if the church ever detaches from rock but doesn't turn off the dag, or if the church has a bunch of prayers (not through rock) and then starts using the dag. I think we could exclude those prayers by adding `where origin_id is not null` to the query above.

tomthornton: @vinnyjth yeah sure, I can add that!

dags/rock/rock_prayer_request_dag.py
tomthornton: I think it was just a bad copy and paste.

vinnyjth (approved):

PR #63, opened 2021-09-08
Added tests

vinnyjth:
I added a readme with instructions, but you can also verify with `pylint` locally! 

There's just a single test in there for now, but now we have a framework to use when testing our app at a broader scale.


tomthornton (approved):
Nice! Worked for me. Thanks for going and figuring out how to do this! I think it will definitely be helpful down the road, especially testing with different churches and how they integrate into the shovel.

PR #64, opened 2021-09-10
Enhancements and fixes for NS

vinnyjth:
Some misc tweaks and fixes I found when working on NS. The issues fixed were...

1. Fixed parent_id being set improperly. The query was correctly identifying the series parent, but then failing to actually use the series parent when updating the `parent_id`
2. Fixed image sizes not being able to bet set correctly because of spaces being improperly encoded. 
3. Fixed issue where an empty key value list feature would crash the features dag. 
4. Stopped paginating in the `content_item` dag if we get an error. We should try fetching those items again instead.


dags/rock/rock_media.py
```diff
@@ -15,7 +15,7 @@
 def getsizes(uri):
     # get file size *and* image size (None if not known)
     try:
-        file = urllib.request.urlopen(uri.replace(" ", ""))
+        file = urllib.request.urlopen(uri.replace(" ", "%20"))
```
tomthornton, line 18: Nice. URI's like this are a pain

dags/rock/rock_content_items_connections.py
```diff
@@ -136,14 +136,15 @@ def run_set_content_item_parent_id(self):
...
+              count(cc.child_id) AS parents_count,
+              p.id AS p_id
               FROM content_item c
               LEFT JOIN content_item_connection cc ON c.id = cc.child_id
               LEFT JOIN content_item p ON p.id = cc.parent_id
               LEFT JOIN content_item_category p_cat ON p.content_item_category_id = p_cat.id
               WHERE p_cat.origin_id IN ({series_parent_category_ids})
-              GROUP BY c.id) AS items_and_parents
-           INNER JOIN content_item_connection ON child_id = items_and_parents.id
+              GROUP BY c.id, p.id) AS items_and_parents
```
tomthornton, line 146: Good catch!

tomthornton (approved):

PR #66, opened 2021-09-11
fixes link and image sanitation

redreceipt:
the second set was overwriting the first

Screen Shot 2021-09-10 at 8 11 12 PM


vinnyjth (approved):
A+ catch

PR #67, opened 2021-09-14
Check for camel cases when linting

vinnyjth:
### About
With this PR, `camelCased` variables will throw linter errors. 

### Testing
Drop a function in the shovel code, anywhere, with a camelcased name. Be sure to `pip3 install pep8-naming` or push the code to the CI to see the error.

### Screenshots
Screen Shot 2021-09-14 at 1 40 53 PM


redreceipt:
forgive me for being the python purist, but why don't we just rename `Klass` to `class_name`, saves a lot of infrastructure requirements and is more python-y

vinnyjth:
> forgive me for being the python purist, but why don't we just rename `Klass` to `class_name`, saves a lot of infrastructure requirements and is more python-y

since `Klass` is still a `Class` instance.  `class_name(kwargs)` feels worse than adding the ignores in my head.

redreceipt:
I just feel like the better solution is figuring out what the standard is instead of adding linter libraries and ignores for every new church DAG

vinnyjth:
@redreceipt The few lines being ignored are all "internal" files and won't be touched by adding new churches. I did some digging and `Klass` as a standin for `Class` didn't seem unusual (although some folks also recommend `cls`) and variables that are instances of a `Class` object seem to be given upperscore names. The linter isn't smart enough to deduce that information.

redreceipt:
PEP8 recommends `class_`

Screen Shot 2021-09-14 at 2 15 46 PM

vinnyjth:
@redreceipt so the change would be to `Class_` which is still a linter error

image

redreceipt:
I like using `kwargs.klass` and `klass(kwargs)` OR `kwargs.class_` and `class_(kwargs)` better than adding linter packages and ignores.

redreceipt:
or perhaps

```py
class ClassName(kwargs["klass"] or None):
    pass

ClassName(kwargs)
```

vinnyjth:
@redreceipt there might be a disconnect here.... the point of this PR is to stop allowing `camelCased` variables without linter errors. I needed to add the `pep8-name` package to accomplish that. Checking `camelCased` variables happens to snag our usage of `Klass`, so I added an ignore for those cases. `Klass` in this case is an instance of a class object. It's not a class instance, or a function. In Ruby it was very idiomatic to reference variables that contained Classes as `Klass` and I've seen that scattered around the python ecosystem as well (although it's certainly not as common as in python.)  

You can poke through Github search results for a bit (https://github.com/search?l=Python&p=8&q=Klass&type=Code) although most of the references aren't directly applicable since they are class instances instead of class literals.

vinnyjth:
> or perhaps
> 
> ```python
> class ClassName(kwargs["klass"] or None):
>     pass
> 
> ClassName(kwargs)
> ```

I'm not sure I have any idea what this is doing

redreceipt:
Ok, I'm with you on the preventing camel-cased variables. 

what about `campus_task = Campus(kwargs) if "klass" not in kwargs else kwargs["klass"](kwargs)`

I just think sprinkling ignores everywhere is a slippery slope and kinda smelly. There may be a reason why class references shouldn't be variables and we can't find any examples of this?

redreceipt (approved):
take or leave my latest comment, I just think ignores are usually an indicator of a better way

vinnyjth:
> campus_task = Campus(kwargs) if "klass" not in kwargs else kwargs["klass"](kwargs)

I'm very much on board with this! 

> I just think sprinkling ignores everywhere is a slippery slope and kinda smelly. There may be a reason why class references shouldn't be variables and we can't find any examples of this?

I was wondering the same thing. I think it definitely a more unique pattern, but I'm not sure it's a smell in itself?

vinnyjth:
I'm going to merge this @redreceipt so it can apply to some PR's currently out there, but I like your approach and will adopt it when I touch these files and when setting up new ones.

PR #68, opened 2021-09-20
adds complete button features

redreceipt:


dags/rock/rock_features.py
```diff
@@ -277,6 +277,23 @@ def get_features(self, content):
                 }
             )
 
+        complete_button_feature = (
+            content["AttributeValues"].get("CompleteButtonText", {}).get("Value")
```
conrad-vanl, line 281: This looks good. On the Rock side for VCC, we'll want to set a default value for this field so they don't have to type it in every time.

conrad-vanl (approved):

PR #70, opened 2021-09-21
fixes bug in updating features

redreceipt:


vinnyjth (approved):

vinnyjth:
@redreceipt Can you fix the tests in this PR? I think they were actually broken by a former PR (the attribute keys one), not sure we are only seeing errors now.

redreceipt:
Yep I’ll get them cleaned up.

redreceipt:
fixed it, needed a `postgres` user since the remote DB we're pulling has it in its schema. Now learning about `vcr` and `casettes` to understand what is happening with this test failure.

vinnyjth:
> fixed it, needed a `postgres` user since the remote DB we're pulling has it in its schema. Now learning about `vcr` and `casettes` to understand what is happening with this test failure.

This might be a case of needing to manually edit the VCR casettes :) The failure appears to be that we are making network requests that are unmocked. However, what has actually changed is the POST data for the network requests. We should be able to solve the issue by adding / editing the YML of the mocks of failing tests so the cassettes we have created are used again.

redreceipt:
ok @vinnyjth take a look

vinnyjth (approved):

PR #71, opened 2021-09-22
feature: Delete Rock content item connections

tomthornton:
### About
Adds tests for content item connections and provides the ability to delete connections from Postgres if removed in Rock

### Testing

`pytest -k "content_item_connections" -s`
### Screenshots
 
Screen Shot 2021-09-22 at 1 38 00 PM


tomthornton:
@vinnyjth ah okay, that makes sense. Need to compare them by parent instead of all at once.

tomthornton:
@vinnyjth latest changes should fix the delta DAG. It matches up by parent id now instead of just plain connection origin id

vinnyjth:
Still seeing the same thing @tomthornton 

Delta DAG is deleting all the connections from my database every time. I think the current approach will always suffer from this issue? Do you think it's worth breaking off the `delete` code into a separate task and then fetching all items from rock and postgres there? 
image

image

tomthornton:
@vinnyjth the problem was this line. Shouldn't have been removing them when they aren't part of the rock response. I removed it and the delta DAG is working now. 
````
else:
     for children_connection_id in children_connection_ids:
          deleted_content_item_connection_ids.append(children_connection_id)```

File: tests/cassettes/content_item_connections/delete_content_item_connection.yaml [hunk 1]
```diff
@@ -0,0 +1,49 @@
...
+      Accept-Encoding:
+      - gzip, deflate
+      Authorization-Token:
+      - ASZjZWdf3IqrbZX9sedtB4wb
+      Connection:
+      - keep-alive
+      User-Agent:
+      - python-requests/2.26.0
+    method: GET
+    uri: https://rock.apollos.app/api/ContentChannelItemAssociations?%24top=10000&%24skip=0&%24select=Id%2CChildContentChannelItemId%2CContentChannelItemId%2COrder&%24orderby=ModifiedDateTime+desc
```
vinnyjth, line 16: I think the reason the tests are passing is that these items wouldn't be returned from Rock if they were deleted. You get nothing back in the real world.

tomthornton: You're right. It worked with one parent item but another when I tried it in Rock.

tomthornton:
@vinnyjth the latest change should actually fix the problem. I think we are going to have to fetch every rock content item connection everytime the delta dag runs, to make sure that they are still there. The only issue I see (that's maybe not even an issue) is that every content item connection is upserted on each dag run. I don't believe this impacts performance?

Rewriting the test now. I'll need to include multiple parent items.

vinnyjth:
@tomthornton We have a pattern in other parts of the shovel of splitting `delete` operations into a seperate dag? What do you think about that? That way we can keep this task `fetch_and_save_content_items_connections` super sharp and focused, and let another task handle the slower operation of checking for deletions.

tomthornton:
@vinnyjth yeah I'm fine with that, it will just mean more requests to Rock.

vinnyjth:
@tomthornton That is totally fine in this case, since there are a few very quick requests made to rock in the original task and then we can leave the potentially slower larger requests for a different task.

dags/rock/rock_content_items_connections.py
```diff
@@ -50,14 +49,61 @@ def map_rock_connection_to_postgres_connection(self, obj):
...
+        )
+        postgres_parent_items = {
+            f"{id}": children for id, children in postgres_parent_item_data
+        }
+        print("Postgres Parent Items")
+        print(postgres_parent_items)
+        print("Rock Parent Items")
+        print(rock_parent_items)
+        deleted_content_item_connection_ids = []
+        for id, children_connection_ids in postgres_parent_items.items():
```
vinnyjth, line 75: Is this methodology (and the associated query) necessary? Is it doing anything differently than the code you had in an earlier commit?

tomthornton: @vinnyjth no it's not doing anything different, but I did add the following so that if a parent item removes it's only child, then it will also be deleted.
```
else:
     for children_connection_id in children_connection_ids:
          deleted_content_item_connection_ids.append(children_connection_id)```

Comment (vinnyjth) [reply dags/rock/rock_content_items_connections.py:75]: @tomthornton Do you think it's worth reverting to the earlier code since it's simpler and easier to follow

**Comment by vinnyjth** [comment]
@tomthornton let me know when you're ready for me to review this :)

**Comment by tomthornton** [comment]
@vinnyjth ready! simplified that deletion logic

File: dags/rock/rock_content_item_dag.py [hunk 3]
```diff
@@ -102,7 +109,7 @@ def create_rock_content_item_dag(church, start_date, schedule_interval, do_backf
 
         media >> set_cover_image
 
-        connections >> set_parent_id >> features
+        connections >> set_parent_id >> features >> delete_connections
```
vinnyjth, line 112: Nice! you can probably have this run concurrently with `set_parent_id` though, is there a reason it needs to wait for `set_parent_id` or `features`?

vinnyjth: Nice! you can probably have this run concurrently with set_parent_id though, is there a reason it needs to wait for set_parent_id or features?

dags/rock/rock_content_items_connections.py
```diff
@@ -54,10 +53,10 @@ def run_fetch_and_save_content_items_connections(self):
         fetched_all = False
         skip = 0
         top = 10000
+        rock_parent_items = {}
```
vinnyjth, line 56: Is this used anywhere?

dags/rock/rock_content_items_connections.py
```diff
@@ -66,9 +65,6 @@ def run_fetch_and_save_content_items_connections(self):
                 "$orderby": "ModifiedDateTime desc",
             }
 
-            if not self.kwargs["do_backfill"]:
```
vinnyjth, line 69: You probably don't mean to delete this.

dags/rock/rock_content_items_connections.py
```diff
@@ -78,6 +74,18 @@ def run_fetch_and_save_content_items_connections(self):
             )
             rock_objects = r.json()
 
+            # Create object of parent items with rock children connection ids
+            for content_item_connection in rock_objects:
```
vinnyjth, line 78: Is this dead code?

dags/rock/rock_content_items_connections.py
```diff
@@ -154,6 +164,69 @@ def run_set_content_item_parent_id(self):
...
+        top = 10000
+        rock_content_item_connection_ids = []
+
+        while not fetched_all:
+            # Fetch content item connections records from Rock.
+            params = {
+                "$top": top,
+                "$skip": skip,
+                # "$expand": "Photo",
+                "$select": "Id,ChildContentChannelItemId,ContentChannelItemId,Order",
```
vinnyjth, line 179: Do you need to fetch the `Order` attribute?

dags/rock/rock_content_items_connections.py
```diff
@@ -154,6 +164,69 @@ def run_set_content_item_parent_id(self):
...
+        postgres_content_item_connection_ids = list(
+            map(
+                lambda x: x[0],
+                self.pg_hook.get_records(
+                    "SELECT origin_id FROM content_item_connection"
+                ),
+            )
+        )
+        deleted_content_item_connection_ids = []
+        print(rock_content_item_connection_ids)
```
vinnyjth, line 218: Do you want to leave this `print` in for the final version of the code?

tomthornton:
@vinnyjth weird I think it deleted your review?

dags/rock/rock_content_items_connections.py
```diff
@@ -179,3 +252,16 @@ def set_content_item_parent_id(ds, *args, **kwargs):
     content_item_connection_task = Klass(kwargs)
 
     content_item_connection_task.run_set_content_item_parent_id()
+
+
+def delete_content_item_connections(ds, *args, **kwargs):
+    if "client" not in kwargs or kwargs["client"] is None:
+        raise Exception("You must configure a client for this operator")
+
+    Klass = (  # noqa N806
```
vinnyjth, line 261: Let's set this task up the way @tomthornton recommended :) I can't remember what it is exactly, but it's involves moving the class instantiation inline. something like 

```
content_item_connection_task = ContentItemConnection(kwargs) if "klass" not in kwargs else kwargs["klass"](kwargs)
```

tests/test_rock_content_item_connections.py
```diff
@@ -0,0 +1,127 @@
...
+import datetime
+from .test_utils import create_test_database, db_connect
+from dags.rock.rock_content_items import ContentItem
+from dags.rock.rock_content_items_connections import ContentItemConnection
+import vcr
+
+create_test_database()
+
+
+def test_run_fetch_and_save_content_item_connections(monkeypatch):
```
vinnyjth, line 11: Nice test! Very easy to follow.

dags/rock/rock_content_item_dag.py
```diff
@@ -102,7 +109,7 @@ def create_rock_content_item_dag(church, start_date, schedule_interval, do_backf
 
         media >> set_cover_image
 
-        connections >> set_parent_id >> features
+        connections >> [delete_connections, set_parent_id] >> features
```
vinnyjth, line 112: is this syntax valid?

dags/rock/rock_content_items_connections.py
```diff
@@ -154,6 +164,69 @@ def run_set_content_item_parent_id(self):
...
+            # Create object of parent items with rock children connection ids
+            for content_item_connection in rock_objects:
+                rock_content_item_connection_ids.append(
+                    str(content_item_connection["Id"])
+                )
+
+            skip += top
+            fetched_all = len(rock_objects) < top
+
+        postgres_content_item_connection_ids = list(
```
vinnyjth, line 209: Nice! I like how this turned out. Simple and easy to follow.

dags/rock/rock_content_items_connections.py
vinnyjth: On second thought, it's not a bad idea since we're paginating.

dags/rock/rock_content_items_connections.py
```diff
@@ -86,6 +94,8 @@ def run_fetch_and_save_content_items_connections(self):
                 skip += top
                 continue
 
+            print(rock_objects)
```
vinnyjth, line 97: dead code?

vinnyjth (changes requested):
Looking good! a few changes needed related to style and dead code / missing code.

dags/rock/rock_content_item_dag.py
vinnyjth: @tomthornton  ^

vinnyjth: Just tested - looks like it is. I thought I had run into issues with this syntax in the past. Apologies!

dags/rock/rock_content_items_connections.py
vinnyjth: Sorry @tomthornton  - I confused myself here :) My question 

> Do you need to fetch the Order attribute?

still stands. I had a seperate question about `$orderby` that I meant to leave my second comment about pagination on.

vinnyjth (approved):
Nice work Tom! Thanks for the great PR description to help get started reviewing.

tomthornton:
@vinnyjth Thanks! The templates were a good idea. Thanks for your help/input on this

PR #73, opened 2021-09-23
feat: Delete Rock inactive people from Postgres

tomthornton:
### About

Rock people who are set to inactive are deleted from the Postgres database

### Testing

Set a user to inactive in rock, run the people DAG, and check that they are deleted from the database

### Screenshots
 

Screen Shot 2021-09-23 at 11 57 04 AM


tomthornton:
@vinnyjth this will need a companion PR in apps that updates the `ON_DELETE` constraint for notification preferences because right now, it causes an error.

tomthornton:
I've also got some merge conflicts to take care of so let's hold off

vinnyjth:
Will try and get this tomorrow. I think this is the right approach for now, we can always "just turn it off" if a church is using the rock "status" field in a way that causes issues.

vinnyjth (approved):
Looking good, thanks tom!

PR #76, opened 2021-10-08
feat: Rock location and event date features

tomthornton:
### About

Adds location and event date features to rock


tomthornton:
> gimme a SQL query that I can test this is working on a core content item

@redreceipt there currently aren't any location fields in core currently. I built this off of VCC. At some point, we'll need @conrad-vanl to add the Location fields to the core rock instance.

dags/rock/rock_features.py
```diff
@@ -434,7 +481,7 @@ def run_fetch_and_save_features(self):
                 "$skip": skip,
                 "loadAttributes": "expanded",
                 "$orderby": "ModifiedDateTime desc",
-                "attributeKeys": "features, comments, buttontext, buttonlink, completeButtonText, scriptures",
+                # "attributeKeys": "features, comments, buttontext, buttonlink, completeButtonText, scriptures, location, eventDate",
```
tomthornton, line 484: This will need to be changed. For some reason EventDate wasn't being pulled in so I commented it out for it to work.

PR #78, opened 2021-10-12
adds VCC duplicate challenge DAG

redreceipt:


dags/vineyard_copy_latest_challenge.py
```diff
@@ -0,0 +1,66 @@
...
+        keepalives=1,
+        keepalives_idle=30,
+        keepalives_interval=10,
+        keepalives_count=5,
+    )
+    last_item_rock_id = pg_hook.get_first(
+        """
+        SELECT origin_id
+        FROM content_item
+        WHERE content_item_category_id = '1acff1cb-805b-4561-b5f7-ee136b0b39b4'
```
vinnyjth, line 22: I think I would order by created at here.

vinnyjth: actually nvm :) that's the rock created at like you pointed at. this is probably fine but maybe a bit unstable ;)

redreceipt:
ok @vinnyjth I think this one is ready

redreceipt:
yup, not sure what I was thinking about

PR #86, opened 2021-11-02
adds woodmen valley setup

redreceipt:


dags/woodmenvalley-rock-content-item-dag.py
```diff
@@ -0,0 +1,139 @@
...
+from rock.rock_cover_image import fetch_and_save_cover_image
+from rock.rock_content_item_categories import (
+    fetch_and_save_content_item_categories,
+    attach_content_item_categories,
+)
+from rock.rock_features import fetch_and_save_features
+from rock.rock_deleted_content_items_dag import remove_deleted_content_items
+
+
+start_date = datetime(2021, 8, 12)
```
vinnyjth, line 19: check your start date ;)

redreceipt: is this start date not ok?

vinnyjth: A start date in the past will cause the delta dag to start on that date and run for every 30 minute period until it catches up with the present date.

redreceipt: oh, nice. since I don't know when this will be merged, I'll just set "catch up" to false

redreceipt: oh ok

vinnyjth: Catching up a day or two is fine, two months might be frustrating.

PR #88, opened 2021-11-02
chore: Content Item Tags from Rock Data Views

tomthornton:
### About
Content Item Data Views are picked up as tags, with a `ContentItem` type

### Testing

run the content item backfill for VCC... you should see `ContentItem` tags and a few  rows in `content_tag`

### Screenshots

`tag` table
Screen Shot 2021-11-02 at 2 07 49 PM

 `content_tag` table
Screen Shot 2021-11-02 at 2 08 18 PM


redreceipt:
DataViews in Rock are for people filters. I think it should just pull from a Tags attribute that is a list of strings.

tomthornton:
Yeah I had it as a multi-select attribute but changed it to DataViews because content attributes are processed in features and we already have tags set up and working with DataViews.

tomthornton:
The UX is also a lot better for data views in rock

redreceipt:
I don't think they can work with content items. But if you can make dataview called "Hope" and tag a content item in core with it proving it works than I'll take a look. I wanna see how it works in the CMS.

redreceipt:
Scratch that :) I see you've already done that! Can you point me to a ragged content item in Rock?

tomthornton:
@redreceipt yeah here's one for VCC https://rock.vineyardcincinnati.com/ContentChannelItem/4156

redreceipt:
@conrad-vanl for the core default I kinda like the DataViews that @tomthornton set up for vineyard. Any reason you think we shouldn't make that the default? and then we can do chase oaks tags custom shovel work?

redreceipt:
going in a different direction

PR #91, opened 2021-11-04
changes to church_slug and removes default columns

redreceipt:


redreceipt:
Would love everyone's blessing on this. I don't understand the implications of this fully but I think it's necessary.

vinnyjth:
This is 100% needed

bkraeling (approved):
Makes sense to me, and the indexes and constraints look how I'd expect them to.

PR #93, opened 2021-11-05
feat: Rock tagging feature to Postgres Content Tags

tomthornton:
### About
- Adds rock `Tags` to postgres as Content Item Tags
- Adds rock content item `DataViews` from rock as Content Item Tags
- Creates `content_tags` in Postgres from `TaggedItems` in rock
- Modifies `deleted_tags` dags to also delete rock `Tags` and rock content item `DataViews`
- Adds `test_rock_tags` test to test the rock `Tags` and rock `TaggedItems` (this test could be better, but for now it just checks that they exist in the database)

### Testing

`pytest -k "rock_tags" -s`

### Screenshots
Adds `Test Tag` and `Another Test Tag`

Screen Shot 2021-11-05 at 11 36 53 AM


tomthornton:
@redreceipt this brings in content item tags from DataViews and Rock's tagging system. I believe @conrad-vanl said both were ideal so I just worked off my other PR from this week.

redreceipt (approved):
💪

PR #97, opened 2021-11-10
feat: Add organization feature for VCC

tomthornton:
### About
Adds an organization feature

### Testing

Run `vineyard_backfill_rock_content_item_dag`

### Screenshot
Screen Shot 2021-11-10 at 12 49 46 PM


nlewis84:
@tomthornton Trying to test this locally, and I get `buildkit not supported by daemon` error when i run `astro dev start`. Have you seen this before?

nlewis84:
Nevermind...I forgot about `DOCKER_BUILDKIT=0 astro dev start`

redreceipt:
@tomthornton this should be custom only for vineyard. Move it over there and I'll approve.

tomthornton:
@redreceipt this can't really be vineyard only. I'd have to make an entire copy of `rock_features` just for vineyard.

redreceipt:
These aren't core features, it needs to be custom. We'll have to figure out a way to make it happen. There shouldn't be unused code in the core libraries.

nlewis84:
@tomthornton I ran the dag successfully, but am not seeing these features added to the feature table. I would expect an additional organization feature to show up here. Some other step that I am missing?

Screen Shot 2021-11-12 at 10 12 06 AM

nlewis84:
Is it an option for us to push up this change, then remove the unused code from the core library sometime next week? Seems like that may be more work than we can finish today.

redreceipt:
The better option would be to copy the whole rock-features file into VCC dags and then shrink it down later

tomthornton:
I believe we could just extend the class and overwrite those one or two functions that need it.

tomthornton:
@redreceipt okay I've added the custom code just for VCC. This should work.

redreceipt (approved):

PR #98, opened 2021-11-17
chore: make tags run as backfill every time

nlewis84:
### About
This PR addresses a problem with the delta DAG properties that was causing rock tag DAG to not work.

Screen Shot 2021-11-17 at 8 39 58 AM

### Testing
@tomthornton I could use some help with testing this out.

### Screenshots

Screen Shot 2021-11-17 at 9 26 18 AM


tomthornton (approved):

vinnyjth:
FWIW - I'm not certain that was actually the issue, it could be something else entirely. I would love to address the root cause :)

nlewis84:
@tomthornton is in there right now checking another issue, so he can probably knock out two birds with one stone.

vinnyjth:
@nlewis84 awesome!

tomthornton:
@vinnyjth yeah I agree. How's this? https://github.com/ApollosProject/apollos-shovel/pull/99

PR #99, opened 2021-11-17
fix: Rock Tags Delta Fix

tomthornton:
### About
This changes the filter params for delta DAG runs for rock tags


vinnyjth (approved):

vinnyjth:
Looks great! Thanks @tomthornton

PR #103, opened 2021-11-23
Setup rock media to support assets

vinnyjth:
Working in Church of Eleven 22

image


redreceipt:
I'm confused on what this does? we already have image URLs in woodmen (and others)

vinnyjth:
Image field types work, but the asset field type storing images does not work currently.

vinnyjth:
I think that's similar to what NS used to store image assets on S3

redreceipt (approved):

PR #105, opened 2021-11-29
chore: Add video name to metadata from Rock

tomthornton:
### About
For selecting different videos within the app, we need names. This will use the abbreviated name or the description

### Testing
Run the content item backfill on core

### Screenshots
Screen Shot 2021-11-29 at 12 26 56 PM


tomthornton:
The rock tag changes are from my code formatter

dags/rock/rock_media.py
```diff
@@ -179,6 +179,8 @@ def map_attributes(self, content_item, attribute):
                     print(image_dimensions)
                     print(attribute)
                     print(media_value)
+        if media_type == "VIDEO" and media_value:
+            metadata["name"] = attribute["AbbreviatedName"] or attribute["Description"]
```
vinnyjth, line 183: Will accessing the attribute dictionary like this result in a `KeyError` if the item doesn't have an "AbbreviatedName " or "Description" key?

tomthornton: Since there is no `$select` param passed, all of the attributes *should* be present. But it couldn't hurt to add the safeget around it.

vinnyjth: True, I'm not sure all rock instances will have the same attributes attached to them out of the box.

vinnyjth (approved):

PR #107, opened 2021-12-01
Added support for action table features

vinnyjth:
### About
Adds support for ActionTable features in the API. 

Right now the interface is pretty strict. 

Your matrix needs to have a JSON formatted value
Your matrix needs at least two keys, "Link" and "Title" for each row.
Your matrix needs to be a attribute with the key "ActionTable"

### Testing

You should see that I've tested it against E22 in the cluster and created an ActionTable feature.

### Screenshots

Screen Shot 2021-12-01 at 3 14 48 PM


redreceipt (approved):
Nice

PR #108, opened 2021-12-09
changes VCC recycle challenge algorithm

redreceipt:


redreceipt:
ok now it's ready, forgot to delete interactions first time around

dags/vineyard_copy_latest_challenge.py
```diff
@@ -1,56 +1,70 @@
...
-        keepalives=1,
-        keepalives_idle=30,
-        keepalives_interval=10,
-        keepalives_count=5,
-    )
+def recycle_challenge():
+    # check if there's an item that starts today
+    headers = {"Authorization-Token": Variable.get("vineyard_rock_token")}
+    items = requests.get(
+        f"{Variable.get('vineyard_rock_api')}/ContentChannelItems?$filter=ContentChannelId eq 2056 and StartDateTime lt datetime'{(date.today() + timedelta(days = 1)).isoformat()}' and StartDateTime gt datetime'{date.today().isoformat()}'",
```
vinnyjth, line 13: you might want to use this utility to make sure that you aren't 4/5 hours off 

https://github.com/ApollosProject/apollos-shovel/blob/master/dags/rock/utilities.py#L132-L144
(it's poorly named)

vinnyjth (approved):
Approving b/c it looks great. Fix the timezone issue and we'll be golden.

dags/vineyard_copy_latest_challenge.py
redreceipt: I don't _think_ it should be an issue because all my comparisons only deal with days with time left off. I'm sending `StartDateTime gt 2021-12-09` and I've confirmed it works. right?

vinnyjth: You would be good if you sent `2021-12-09` but you're sending `isoformat` which includes a timestamp

vinnyjth: OH NVM!

redreceipt: no it doesn't
Screen Shot 2021-12-09 at 3 35 33 PM

redreceipt: `datetime` would

vinnyjth: Yep. I saw the same thing. You're good.

PR #109, opened 2021-12-16
feat: People connections

tomthornton:
### About
This PR adds a `fetch_and_save_people_connections` dag which adds `people_connection` rows based on content item attributes that are a Person field type.
### Testing
First, you'll need [this PR](https://github.com/ApollosProject/apollos-apps/pull/2296) from core to create the `people_connection` table. You'll need to run `yarn migrator up` in the `ApollosApi` folder.

Second, you'll need to run the `core_backfill_people_rock_dag`. Once that's complete, you should see the record below in the database.

### Screenshots
Screen Shot 2021-12-16 at 1 11 36 PM


vinnyjth:
Sorry @tomthornton, I'm out of the loop. What's the context for this PR?

tomthornton:
@vinnyjth https://github.com/ApollosProject/apollos-apps/issues/2208 We're trying to expose people connections from Rock in the app

vinnyjth:
@tomthornton Do you think this would be better exposed as a feature that could be on a content item? 

On the shovel side, that would look like identifying people for a piece of content, and then making a feature with `data` containing a list of people? 

My primary concern with this aproach (which is very well built otherwise :) is that it's a bit generic. I can't tell what a person connection is, what it is used for, or how it correlates to data in a CMS.  Is it people who like content? People who have visited a piece of content? People who are preaching in that content? etc.

tomthornton:
@vinnyjth yeah the idea was to make it more flexible so that the attributes on the content items could be named in the way that the person relates to the content, whether that be a speaker, preacher, content author... etc. A person connection is simply an attribute on a content item with the field type `Person`.

As far as it being a feature... I'm not sure how we could have multiple people with descriptions as a field in Rock. 

Also, I tried to go the route of just fetching `AttributeValues` that had a `FieldTypeId` of the type `Person` but it didn't work. Couldn't get the right data.

vinnyjth:
> @tomthornton yeah the idea was to make it more flexible so that the attributes on the content items could be named in the way that the person relates to the content, whether that be a speaker, preacher, content author... etc. A person connection is simply an attribute on a content item with the field type Person.

Can you explain this more? I can see the flexibility on the Rock Side, but on our API side it looks like we will have no way of knowing (and displaying) what type of connection it is (author, preacher, speaker, etc) . That might not be a huge issue? 

Imagining that we lived in a perfect world and we could do whatever we wanted from a data fetching standpoint, do you think this is better represented as a feature or a content item attribute?

vinnyjth:
here's an example of fetching attributes of a specific type

https://github.com/ApollosProject/apollos-shovel/blob/master/dags/rock/rock_tags.py#L413-L429

People will be tricker since the `value` will likely be a `guid` you can then use that guid to find the correct people like `GET /api/People?$filter=Guid eq guid${guide}`

tomthornton:
> here's an example of fetching attributes of a specific type
> 
> https://github.com/ApollosProject/apollos-shovel/blob/master/dags/rock/rock_tags.py#L413-L429
> 
> People will be tricker since the `value` will likely be a `guid` you can then use that guid to find the correct people like `GET /api/People?$filter=Guid eq guid${guide}`

@vinnyjth hmm i tried that approach and it didn't work :/ I can try it again though once we figure out the approach

tomthornton:
> > @tomthornton yeah the idea was to make it more flexible so that the attributes on the content items could be named in the way that the person relates to the content, whether that be a speaker, preacher, content author... etc. A person connection is simply an attribute on a content item with the field type Person.
> 
> Can you explain this more? I can see the flexibility on the Rock Side, but on our API side it looks like we will have no way of knowing (and displaying) what type of connection it is (author, preacher, speaker, etc) . That might not be a huge issue?
> 
> Imagining that we lived in a perfect world and we could do whatever we wanted from a data fetching standpoint, do you think this is better represented as a feature or a content item attribute?

@vinnyjth so I've already built the UI portion of this. (I had both API changes and UI changes in one big PR but thought it might be best to split them up in order of when they were merged). I can add that as a draft PR if you wanna take a look. Shovel uses the attribute name to assign a type, which is rendered in `FollowListItem` as a description. 

As a feature, could we still attach Rock People? I just feel like the Person content item attribute is a really easy and customizable way for this to work. Maybe I'm not fully understanding what you mean though.

vinnyjth:
Gotcha, thanks for the context @tomthornton! 

@conrad-vanl What do you think? Should we build out this "people" block as a Feature or as a ContentItem attribute?

conrad-vanl:
> but on our API side it looks like we will have no way of knowing (and displaying) what type of connection it is (author, preacher, speaker, etc)

I think this is important. From a data model perspective, I think we need something to represent the "Edges" - thinking of a graph, the line that's attaching a content item to a person. What does that edge represent? Author, Preacher, Speaker, etc.

> "people" block as a Feature or as a ContentItem attribute?

Not sure. I might lean towards the latter, to be honest. We need to talk more about the future of Features - they're great for representing custom content structures - but they can be hard to reason about and understand since they're so abstract - and so I could argue that for built-in things like Videos - those should never be features? But I'm a little 🤷‍♂️ on that approach. 

This task was a bit of a pie in the sky task and surprised (but happy) @tomthornton got to working on it 😀

conrad-vanl:
I think the important things to have:

- To be able to display people connected to a node (maybe we need like a PeopleConnectionNode or something if it's not feature based????)
- To be able to "follow" these people in the same way you can follow any other user
- To be able to trigger a notification when there's a new content item in the database connected to a person you follow

Use case:
I follow a specific preacher at my church. I get a push notification when he has a new sermon to watch.

vinnyjth:
@conrad-vanl right on - that's helpful

@tomthornton In that context, I'm on board with the data model of this PR but I have feedback relevant to the GQL portion in apps that I'll move there :) 

However, I think this PR is missing `people_connection` deletions. If you remove a person from a content item, that change won't be reflected in the database.

tomthornton:
@conrad-vanl to your first point, the `type` column in the `people_connection` table should designate the type of connection it is. I originally had it as an enum, but ultimately changed it to a string to allow some flexibility.

@vinnyjth good call! I'll add that tomorrow.

PR #114, opened 2022-01-17
Contentful Sync (Kids Club Data Only)

liamillard:
### About

This PR adds the sync capability to our contentful integration and creates the sync for the Crossroads Kids Club Data.

We're now saving sync tokens in redis. Kids club sync tokens are saved under the key `crossroads_kids_club_contentful_sync_token`. I did not use the redis hook that airflow built because it's surprisingly limited. Instead, we're using the regular ol' `redis` python package. The redis connection URL is now stored as an airflow variable.

To get the sync working, I had to tweak the original backfill. We had been backfilling video content only, but that led to issues during the sync. When you specify a content type (in this case, we specified that we were backfilling `Entries` of type `Video`), all following syncs also only use that specific type. In order to get updates to assets and categories (in the future), I had to open up the backfill to sync _all_ content, not just entries of type `Video`.

I definitely wasn't able to test all of the sync scenarios, so we'll watch this run over the next few days to make sure its behaving as expected. 

### Testing

1. Start up Airflow.
2. Toggle on the `crossroads_kids_club_backfill_contentful_content_item_dag` DAG. Wait for it to finish, it will take a few minutes.
3. Toggle on the `crossroads_kids_club_contentful_content_item_dag`, and wait for it to finish. It should run much faster 😅

### Screenshots
 
Screen Shot 2022-01-17 at 6 07 10 PM


dags/contentful_crossroads/cr_redis.py
```diff
@@ -0,0 +1,6 @@
+import redis
+from airflow.models import Variable
+
+redis_client = redis.Redis.from_url(Variable.get("redis"))
```
vinnyjth, line 4: What do you think about prefixing your variable with `crossroads_kids_club`? For the event that we ever use Redis for something else :)

liamillard: I decided not to prefix this airflow variable, thinking that if another church uses redis they'll use the same connection string. But anything I write to redis is prefixed with the `crossroads_kids_club`. Does that make sense? Or will different churches use different redis connection strings?

vinnyjth: Yah, I see the logic :) I'm thinking about instances where we potentially integrate with someone else's Redis, or if we support a church with their own cluster. Do you think we would be better off to namespace for now, and then setup a global if we need it?

liamillard: Oh I see - totally, I'll go ahead and add the prefix 👍

dags/crossroads_kids_club_content_item_dag.py
```diff
@@ -15,7 +15,7 @@
 globals()[backfill_name] = backfill_dag
 
 dag, name = create_cr_kids_club_content_item_dag(
-    start_date, timedelta(minutes=30), False
+    start_date, timedelta(minutes=15), False
```
vinnyjth, line 18: One thing I noticed while testing locally (I missed this the first time around) 

You'll want to set your start date to a couple days ago. Right now it's going to try and run this for every single fifteen minute window that's existed from November 17th -> Now :)

liamillard: 😱 YIKES that is definitely not what we want haha. I'll update that too!

liamillard:
This should be ready to go again @vinnyjth!

vinnyjth (approved):

PR #117, opened 2022-01-20
fix: First item not being deleted from Crossroads Content DAG

vinnyjth:
### About
I think I found a bug, the first id in the array of items to delete has a `$` prefixed, so it will never find and delete that item :)


liamillard (approved):
Ah good catch! This is me trying to write javascript in a python script 😆

PR #118, opened 2022-01-21
fixes prayer DAGs

redreceipt:
this will skip prayers from people that are inactive and also skip prayers that are private


redreceipt:
@vinnyjth this is unrelated to the other issue we discussed but this is a necessary step one

dags/rock/rock_prayer_requests.py
```diff
@@ -19,24 +19,20 @@ def __init__(self, kwargs):
...
-            (str(rock_person_origin_id),),
-        )
-
-        return postgres_person_id[0]
-
     # "created_at", "updated_at", "text", "person_id", "origin_id", "origin_type", "apollos_type"
     def map_prayer_requests(self, prayer_request):
-        postgres_person_id = self.get_postgres_person_id(
-            safeget(prayer_request, "RequestedByPersonAlias", "PersonId")
+        rock_id = prayer_request["RequestedByPersonAlias"]["PersonId"]
```
vinnyjth, line 24: you might want to safeget this, I think we saw some prayers that had no `RequestedByPersonAlias` object (because they were artificially inserted, not created by a user)

dags/rock/rock_prayer_requests.py
```diff
@@ -78,13 +74,25 @@ def run_fetch_and_save_prayer_requests(self):
...
+            ]
+            # filter out skipped prayers
+            prayer_requests = [prayer for prayer in prayer_requests if prayer]
 
             data_to_insert, columns, constraints = find_supported_fields(
                 pg_hook=self.pg_hook,
                 table_name="prayer_request",
                 insert_data=prayer_requests,
             )
+            print("constraints", constraints)
```
redreceipt, line 95:
```suggestion
```

dags/rock/rock_prayer_requests.py
redreceipt: it should bomb out if it gets here with no ID, I've insured that shouldn't happen one level up

vinnyjth (approved):

camrun91:
sweet I will pul this into our shovel and see how it goes.

camrun91:
@redreceipt So I just thought of a scenario that we see fairly often that I am not sure this accounts for. 
A prayer is submitted via website and put in as public. Then our prayer team sees it actually does contain non-public info. If they mark the prayer as non public in Rock this would not get reflected in the shovel would it?

redreceipt:
only if the delete prayers DAG catches it. not sure. if you see otherwise, let us know.

PR #119, opened 2022-01-28
Crossroads Kids Club - Contentful Refactor

liamillard:
### About

I found myself wanting to change _a lot_ of things in the Contentful backfill and sync process after really digging into the Crossroads TV content. Also, I actually read some of their [docs on the Contentful python package](https://contentful.github.io/contentful.py/), and it's way more capable than I was giving it credit for.

Changes I made:
1. There's no more kids-club-specific stuff in `crossroads_contentful/cr_contentful_entries.py`. All of the kids' club functionality is defined in the new `CrossroadsKidsClubContentItem` class, which is built off the `ContentItem` class.
2. We are no longer syncing assets for the first step! The contentful python classes have a `.resolve()` function that allows you to resolve links to other assets/entries. So we're using that instead 🎉
3. I was pretty hackily accessing members on the entries and assets. I cleaned that up after reading the docs. So there's no more `entry.sys["id"]`, because I can access that directly using `entry.id`. All fields are also members on the `entry`, so there's no more use of `_fields` or `get_fields_for_entry`. It's just cleaner.
4. During the backfill and sync, I'm specifying the `type: Entry`. This allows me to further specify that only `videos` are synced for kids club, which _drastically_ speeds up the backfill. One catch: deleted items will not show up in future syncs, so we have to subscribe to a Deletion sync separately.

> For entries, you can also specify a content_type parameter. When specifying content_type you must specify type as Entry (meaning there will be no deletions). You can only specify the type and content_type parameter at the initial sync along with the initial parameter. Any subsequent syncs will only include the types you have specified. If you want to sync entries by content type you should separately subscribe to a `Deletion` or `DeletedEntry` sync to be notified when entries are deleted. ([Source](https://www.contentful.com/developers/docs/references/content-delivery-api/#/reference/synchronization/initial-synchronization))


liamillard:
@vinnyjth this isn't blocking me yet - I'm going to just write the Crossroads TV stuff on a branch of this branch,  then cherry-pick the commits later. Also I'm not sure what's going on with the `black check` step in the lint test, it works well for me locally 🤔

vinnyjth:
@liamillard this looks like the black error. very weird! 

```
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.10.2/x64/bin/black", line 5, in <module>
    from black import patched_main
  File "src/black/__init__.py", line 35, in <module>
ImportError: cannot import name 'ParameterSource' from 'click.core' (/opt/hostedtoolcache/Python/3.10.2/x64/lib/python3.10/site-packages/click/core.py)
Error: Process completed with exit code 1.
```

I'll see if I can solve that :)

dags/contentful_crossroads/kids_club/cr_kids_club_content_item_dag.py
```diff
@@ -46,42 +47,31 @@ def create_cr_kids_club_content_item_dag(
         ),
     )
 
-    sync_token = redis_client.get(kc_redis_token)
-
-    if do_backfill or (not do_backfill and not sync_token):
+    redis = RedisClient(client=church)
+    entry_sync_token = redis.get_sync_token("entry")
```
vinnyjth, line 51: We can move this to a seperate PR, but I just realized that we shouldn't be fetching the sync token in this context 😬  Essentially this `redis` line gets called a couple times a second since it's included in the top level

https://airflow.apache.org/docs/apache-airflow/stable/best-practices.html#top-level-python-code

I might refactor so you..

1. keep backfill separate like before
2. conditionally / don't backfill depending on the presence of the sync token inside the non-backfill  

this might of been the way things have been prior? 

It sucks, but I don't think it's possible to know if you have a sync token before firing off `fetch_and_save_entries` without breaking airflow's constraints.

vinnyjth: Let me know if you would rather fix this here, or fix this in a future PR :) I'm fine either way. 

It won't break airflow/redis (and it's already doing this in prod currently) but it's def a performance concern long term.

vinnyjth (approved):

liamillard:
I found one more issue - about to push a fix

PR #120, opened 2022-02-07
Crossroads TV Backfill and Sync (via Contentful)

liamillard:
### About

This PR adds the backfill and sync for the Crossroads TV content.

### Backfill:
Screen Shot 2022-02-07 at 12 10 19 PM

#### `create_categories`
In this step, we're creating 3 hard-coded categories for the Crossroads TV content: In-Room Messages, Messages, and Videos.

#### `fetch_and_save_entries`
In this step we retrieve entries from contentful and save them to postgres. At the same time, we'll save the asset associated with an entry, save an item to the media table for it's video URL, and link it to the correct category.

#### `fetch_and_delete_items`
Because the step above is specific for entries, we also have to subscribe to a `deletion` sync. This step handles entries and assets that were deleted from contentful.

#### `link_entries`
This step links parent content (collections or series) to their children. Each child in `content_items` is updated with a `parent_id`, and a row is added to the `content_item_connections` table.

### Sync:
Screen Shot 2022-02-07 at 12 20 03 PM

The sync process does not include `create_categories`, but is otherwise the same as the backfill. We're storing the sync tokens in redis, and only content added or edited since the last sync token is synced.

### Testing

1. Start up Airflow, and filter the dags list by `crossroads_tv`
2. Unpause the `crossroads_tv_backfill_contentful_content_item_dag`. Wait for it to complete, it usually takes >20 minutes. (If you don't want to wait for it to complete, you can check out the crossroads_tv staging postgres DB and look at the data from the last time I ran the DAG.
3. Unpause the `crossroads_tv_contentful_content_item_dag`. Wait for it to finish. It should only take a few minutes.
4. Check out all the data in postgres 😅

### Screenshots
 

|Backfill|Sync|
|---|---|
|Screen Shot 2022-02-07 at 12 23 34 PM|Screen Shot 2022-02-07 at 12 23 22 PM|


vinnyjth:
I love your PR description 😍  thanks for putting so much thought into this!

dags/contentful_crossroads/cr_contentful_series_collections.py
```diff
@@ -0,0 +1,145 @@
...
+        self.pg_hook.insert_rows(
+            "content_item_connection",
+            content_to_insert,
+            columns,
+            0,
+            True,
+            replace_index=constraint,
+        )
+
+    def add_apollos_ids(self):
```
vinnyjth, line 107: I need to find a DB solution for this - it's kinda annoying to have to do this in all your new shovels :)

dags/contentful_crossroads/tv/cr_tv_content_item_dag.py
```diff
@@ -0,0 +1,123 @@
...
+        tags=tags,
+        # Let tasks run no more than three times longer than the schedule interval.
+        dagrun_timeout=(
+            schedule_interval * 3 if type(schedule_interval) is not str else None
+        ),
+    )
+
+    if do_backfill:
+        with dag:
+            create_categories = PythonOperator(
```
vinnyjth, line 53: Is there any reason you can't run this w/ the non_backfill as well? 

I only ask because it might reduce the amount of branches in this file, you could probably eliminate the `do_backfill` if/else :)

vinnyjth (approved):

dags/contentful_crossroads/tv/cr_tv_content_item_dag.py
liamillard: Yeah you're right, we could totally run the `create_categories` step during a sync too. It shouldn't hurt anything because the unique fields prevent me from duplicating things. And it runs SUPER fast.

PR #121, opened 2022-02-07
Crossroads Algolia Indexes

liamillard:
### About

This PR creates an algolia index for Crossroads Kids Club and adds the indexing step to both the backfill dag and the sync dag.

I largely copied the code from `dags/algolia_multitenant/algolia_content_dag.py`, and made changes necessary for crossroads. The following changes are relevant:

- We're using Crossroads' Algolia instance instead of the cluster's. Crossroads pays for a fancy algolia account and I see us wanting to use some of those features in the future. 
- We're using the description for the summary field in algolia. To do so, I had to remove the HTML markup from the description field using the new `HTMLParser` class.

### Testing
1. Run the backfill DAG for kids club. Verify that the algolia indexing step completes.
2. Check out the results in algolia's dashboard (I'm happy to give you the credentials for this account) and verify everything looks good.

### Screenshots
|Backfill|Sync|
|---|---|
|Screen Shot 2022-02-07 at 3 07 36 PM|Screen Shot 2022-02-07 at 2 58 06 PM|


liamillard:
@vinnyjth I think to save review time of yours, I'll wait until #120 gets in, then I'll add the Crossroads TV algolia indexing to this PR as well. That should only be a few more additional lines changed.

PR #122, opened 2022-02-08
Crossroads Algolia Indexing

liamillard:
### About

This PR creates an algolia index for Crossroads Kids Club and Crossroads TV and adds the indexing step to both the backfill dag and the sync dag.

I largely copied the code from `dags/algolia_multitenant/algolia_content_dag.py`, and made changes necessary for crossroads. The following changes are relevant:

- We're using Crossroads' Algolia instance instead of the cluster's. Crossroads pays for a fancy algolia account and I see us wanting to use some of those features in the future. 
- I'm mapping the description field to the summary field in algolia. Crossroads content only has a description, no summary. To do so, I had to remove the HTML markup from the description field using the new `HTMLParser` class.

I went ahead and took this opportunity to reorganize the DAGs a bit too. For both kids club and tv, we had only been running the `create_categories` step during a backfill. There's no reason we can't run the `create_categories` step during a sync too, so I removed some unnecessary logic from each of the DAGs.

### Testing
1. Run the sync DAG for kids club and crossroads tv. Verify that the algolia indexing step completes.
2. Check out the results in algolia's dashboard (I'm happy to give you the credentials for this account) and verify everything looks good.

### Screenshots
||Backfill|Sync|
|---|---|---|
|Kids' Club|Screen Shot 2022-02-07 at 3 07 36 PM|Screen Shot 2022-02-07 at 2 58 06 PM|
|Crossroads TV|Screen Shot 2022-02-09 at 10 03 19 AM|Screen Shot 2022-02-09 at 10 02 59 AM|


vinnyjth:
@liamillard Thanks for handling the merge conflicts! I was going to offer to do them for you since I'm sure I made a big mess, but seems like you knocked it out :D

vinnyjth (approved):

PR #124, opened 2022-02-09
Dag reorg

vinnyjth:
### About
I was starting to get frustrated with how unorganized our `dag` folder was becoming. All the files have been shuffled around into one of three locations 

`dags` still live on the root. This is the actual dag creation functions

Individual uses of a dag - willowcreek's dags for example - now live in `churches/willowcreek/*`

The operators (tasks) of a dag now live in an `operators` folder. 

This should get us a lot closer to a sustainable structural ecosystem. I also dropped a file called `ORGNIZATION.md` into the folder to serve as a living document of sorts. 

### Screenshots
image

image


vinnyjth:
@liamillard Don't feel obligated to review this, but I wanted to get your eyes on it since you've probably got a really good sense of what is wrong with our current organization structure :)

redreceipt (approved):
Love it

PR #127, opened 2022-02-21
Crossroads Kids' Club Content Updates

liamillard:
### About
This PR updates what kids club content we save:

- Crossroads has implemented a new distribution channel for kids club content, so we're only saving content with the `Kids Club App/TV` distribution channel
- We're no longer checking for the kids club category before saving. We're relying on the distribution channel exclusively to dictate what content should be saved.
- We're checking for a streaming URL before saving content too, as that's required for OTT
- We added support for 2 new categories: Gods Story, and Kids' Club Music. These are still largely hardcoded. Until they make official categories for the "For Older Kids" and "For Younger Kids" videos, these categories will stay hardcoded.

### Testing
1. Rerun a backfill and sync and verify that all still work 👍


dags/operators/contentful_crossroads/cr_contentful_content_items.py
```diff
@@ -239,7 +239,7 @@ def get_category_id(self, entry):
         pg_entry_id = None
         try:
             pg_entry_id = self.pg_hook.get_first(
-                f"SELECT id FROM content_item_category WHERE origin_id = {entry.category.id}"
+                f"SELECT id FROM content_item_category WHERE origin_id = '{entry.category.id}'"
```
liamillard, line 242: This is a bug I found while looking up kids club categories, so I thought I'd fix it here too though it's unrelated to this PR.

dags/operators/utils/database.py
```diff
@@ -33,7 +33,7 @@ def get_conn(self) -> connection:
             password=conn.password,
             dbname=self.schema or conn.schema,
             port=conn.port,
-            **pg_hook_defaults # Pull in our default connection params :apollos
+            **pg_hook_defaults  # Pull in our default connection params :apollos
```
vinnyjth, line 36: thanks :D

vinnyjth (approved):

PR #128, opened 2022-02-23
Kids Club Categories Update

liamillard:
### About
Crossroads decided to change how they handle two of the 4 categories we support: Older Kids and Younger Kids. They each have their own category in contentful now, so we no longer have to parse the titles of videos to determine which category a video should be assigned to. All good things 👍

This PR updates the handling of categories to pull them from contentful, then save them to postgres. Then we simply assign a content item to its category based on the category id.

It's worth mentioning that I'll have to manually delete the old categories from the production database after this PR is merged and the backfill runs successfully. I was unable to do this programmatically because some content items are still assigned to the old hardcoded categories. That is until the backfill runs again.

### Testing
Smoke test the backfill. It should work.


dags/operators/contentful_crossroads/kids_club/cr_kids_club_categories.py
```diff
@@ -1,48 +1,38 @@
...
-            "origin_id": origin_id,
+            "origin_id": category.id,
             "origin_type": "contentful",
             "apollos_type": "ContentChannel",
-            "title": category["title"],
+            "title": category.title,
         }
 
     def run_fetch_and_save_categories(self):
         # There's no fetching involved yet.
```
vinnyjth, line 29:
```suggestion
        # There's now fetching ;)
```

PR #134, opened 2022-03-16
adds campus tags

redreceipt:


redreceipt:
@vinnyjth I'm gonna do the campus tags incrementally. The only thing I don't like about this is setting `origin_id` to the rock campus ID. it's misleading when looking at the DB and doesn't tell you how to get back to the origin. how would you approach it? only option I see is setting the `origin_type=rock_campus` or something but that's also an anti-pattern

conrad-vanl:
@vinnyjth I'd like to bump your 👀 on this as it is currently blocking Willow Creek. 

@redreceipt I'd really like to get this algorithm needs for WC patched up this week. I believe that's having a campaigns feed that takes into account both personas and campuses, which this is the start of.

redreceipt:
Yep. My plan is just to repurpose the deprecated campaigns algo to do exactly that

dags/operators/rock/rock_tags.py
```diff
@@ -536,6 +522,62 @@ def run_attach_persona_tags_to_content(self):
...
+            SELECT name, origin_id
+            FROM campus
+            """
+        )
+
+        campuses = [
+            {
+                "created_at": self.kwargs["execution_date"],
+                "updated_at": self.kwargs["execution_date"],
+                "origin_id": campus[1],
```
vinnyjth, line 538: for instances like this - we've done `f"campus-{campus[1]}"` to setup the association correctly

vinnyjth:
@redreceipt are you planning on doing a followup PR to attach campus tags to people?

conrad-vanl:
I'm confused....how are we doing campus tags currently? I know we have things filtered by campus on the Woodmen App?

dags/operators/rock/rock_tags.py
```diff
@@ -192,30 +180,33 @@ def run_fetch_and_save_rock_tags(self):
...
 
             skip += top
-            fetched_all = len(rock_objects) < top
-
-            tags_to_insert = list(map(self.map_rock_tag_to_tag, rock_objects))
+            fetched_all = len(rock_tags) < top
+
+            tags = [
+                {
+                    "created_at": self.kwargs["execution_date"],
```
vinnyjth, line 194: I think it's better to have these broken out into discrete methods, so they can be overridden by subclasses.  This is pretty common on some of the other operators.

conrad-vanl:
I'm confused by @vinnyjth 's question:

> @redreceipt are you planning on doing a followup PR to attach campus tags to people?

vinnyjth:
> @vinnyjth We won’t need campus tags to people but will need campus tags to content. Yes that’ll be in the next one. So originID can just be an arbitrary string? Feels like a smell to have it be different things in the same table.

It's not arbitrary - it's a distinct pointer at a specific remote record ;)

redreceipt:
> > Copy and pasting the campus name into the rock generic tags
> 
> 
> 
> Yeah but how is the person getting added to the tag?

People already have campuses attached to them. Don't need tags.

vinnyjth:
@conrad-vanl @redreceipt if you get a chance, maybe you could pull together the complete strategy for me and note how this PR is related? 

Here's what I got- 

1. We now have campus tags in the db
2. We are going add relationships between campus tags and items
3. The client will ask for feed for a given campus **name**
4. Given the name of the campus, we will look for items with a name matching that tag?

vinnyjth:
https://github.com/ApollosProject/apollos-cluster/blob/10c50546ef712147cf3986a88cefee1bfa3453e7/packages/apollos-data-connector-postgres/src/action-algorithms/index.js#L214

More context ^

redreceipt:
> @conrad-vanl @redreceipt if you get a chance, maybe you could pull together the complete strategy for me and note how this PR is related?
> 
> Here's what I got-
> 
> 1. We now have campus tags in the db
> 2. We are going add relationships between campus tags and items
> 3. The client will ask for feed for a given campus **name**
> 4. Given the name of the campus, we will look for items with a name matching that tag?

I was assuming we'd lookup campus by context, not query arg. Like we do with content feed in your reference above. But everything else is correct. Is there a piece that is still unclear. I can write a post or we can chat on a call real quick?

dags/operators/rock/rock_tags.py
redreceipt: I just don't see how it could be mapped any other way for a custom purpose. and we don't have the possibility for custom DBs anymore. having a function that is only referenced in one place is an unneeded abstraction that makes the code hard to read and not to mention python comprehensions are a bit easier to understand, wouldn't you agree?

dags/operators/rock/rock_tags.py
redreceipt: I can do this. just didn't know if it might cause problems later trying to bulk operate on tags with origin IDs looking different

dags/operators/rock/rock_tags.py
vinnyjth: I don't :) I find the mapping in a distinct function easier to parse. It isolates the complexity a bit more, and allows me to scan the operator a bit better. 

We don't have a option for custom DBs, but we do have an option for custom rock instances (which is where those custom mappers come into play)

dags/operators/rock/rock_tags.py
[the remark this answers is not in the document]
vinnyjth: ah thanks!

dags/operators/rock/rock_tags.py
```diff
@@ -536,6 +522,62 @@ def run_attach_persona_tags_to_content(self):
...
+
+        campuses = [
+            {
+                "created_at": self.kwargs["execution_date"],
+                "updated_at": self.kwargs["execution_date"],
+                "origin_id": campus[1],
+                "origin_type": "rock",
+                "apollos_type": "Tag",
+                "name": campus[0],
+                "data": None,
```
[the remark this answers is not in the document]
vinnyjth: right. tags are a bit weird because they have a higher degree of potential polymorphism. We could make the data model a lot simpler if we attached people to `campus` tags

redreceipt:
ok @vinnyjth I put the mapper back and added a custom origin ID

vinnyjth (approved):

PR #135, opened 2022-03-17
Groups

camrun91:
### About
Adds group Dag 

### Testing
not ready to test yet. But eventually you should be able to add the groups job to the core dag


camrun91:
@vinnyjth I have not got the airflow stuff set up on my mac yet. But I wanted to make sure I pushed all this up and got eyes on it.

One thing I am not clear on is the different ways we call the PG_hook to run. In some dags we call it in this way in others we call run directly. If there is documentation I need to read up on to fully understand I can do that. If it is more of a matter that we just have not been consistent that is completely understandable as well.

camrun91:
Here is the corresponding draft pr in cluster https://github.com/ApollosProject/apollos-cluster/pull/128

I am stricly focusing on getting the groups to sync

camrun91:
This is almost ready it is working with one GroupType. However, I am trying to figure out how to best cycle through more than one group type. Just need to add them to the filter param in the Rock query @redreceipt @vinnyjth got any ideas or examples?

camrun91:
@vinnyjth this should have the "Core" functionality to bring in Groups with the ability to add specific Type ID's.

vinnyjth:
@camrun91 run `black` locally and commit the results :)

camrun91:
@vinnyjth ok this should be good now.

PR #136, opened 2022-03-21
chore: added rock people and rock tags dags for City First and Oak Cliff

nlewis84:


dags/churches/city_first/city-first-rock-people-dag.py
```diff
@@ -0,0 +1,19 @@
+from airflow import DAG  # noqa: F401
+
+from datetime import datetime, timedelta
+
+from rock.rock_people_dag import create_rock_people_dag
```
vinnyjth, line 5: this import is wrong I think :)

nlewis84:
@vinnyjth I wasn't totally sure on the best way to customize the rock_tag_dag, so I just made a new one that is specific for only people_tags. I am getting a None error currently that I'm not sure about, but may be related to the fact that we do not have a Rock Token yet, and I just defined that variable as an empty string.

vinnyjth:
@nlewis84 Looks great to me. Once we get the Rock tokens I'll test and merge this.

vinnyjth:
@nlewis84 let me know when we have rock variables for both churches

vinnyjth (approved):
💥

PR #139, opened 2022-03-23
attach campus tags to content items

redreceipt:


dags/operators/rock/rock_tags.py
```diff
@@ -569,6 +569,95 @@ def add_campus_tags(self):
...
+                WHERE origin_id = '{rock_content_id}'
+                """
+            )[0]
+
+            # delete any existing campus tags
+            self.pg_hook.run(
+                f"""
+                DELETE
+                FROM content_tag
+                WHERE content_item_id = '{content_item_id}'
```
redreceipt, line 648: I need to limit this to only campus tags

redreceipt:
@vinnnyjth this one is ready

dags/operators/rock/rock_tags.py
```diff
@@ -569,6 +570,96 @@ def add_campus_tags(self):
...
+            content_item_id = self.pg_hook.get_first(
+                f"""
+                SELECT id
+                FROM content_item
+                WHERE origin_id = '{rock_content_id}'
+                """
+            )[0]
+
+            # delete any existing campus tags
+            campus_tag_ids = str([tag[0] for tag in campus_tags]).strip("[]")
```
vinnyjth, line 643: I like that you've got deletion support built in 👍

vinnyjth: have you considered excluding the campus_tags you already have from deletion, and then doing an `upsert` on the campus tags that you do have?

redreceipt: I did consider that but since the performance seemed almost the same, I opted for the less complicated code logic. But I’m open to changing if you think it would be better?

PR #140, opened 2022-03-24
Fix shortcodes; sped up the content dag a bit

vinnyjth:


vinnyjth:
@nlewis84 made a PR so you can track this a bit better, I'm going to go ahead and merge though :)

PR #142, opened 2022-03-30
makes button depend on link instead of text

redreceipt:


nlewis84:
Oh, I see. So we are checking for the actual URL instead of the text, in case the text is blank, and then also setting the text to "Go!" if it is blank. Nice!

